PR #225, opened 2015-08-12
Create Folder model, model spec, factory, serializer, endpoint, and endpoint spec

JamesM967:
Could probably use some refactoring here and there. I'll probably just close the waffle endpoint cards when DM/DL approve. Also the Parent/Child Folder endpoint hasn't been created yet. I can work on that next


JamesM967:
Leave this 'in progress' until the pull request is actually merged

PR #517, opened 2016-03-15
Merge latest api docs into official upstream develop

mgardnerpsu:


dmlond:
merged into dmlond fork via commandline, so this merge will be closed without merge

PR #529, opened 2016-03-17
added ability to enter an existing api_token (e.g. for a software_agent) to swagger authentication

dmlond:


dmlond:
oops, broke the build :)

PR #623, opened 2016-05-09
Adding Polymorphic Association

jonathanparrish:
Things to remember:

Tag.unscoped.where(taggable: DataFileScope::Scope.new(user, DataFile.all.reorder(''))).group(:label).count

from our email


dmann:
Ended up implementing this a different way.

PR #628, opened 2016-05-18
Only call graphed methods in the specs that are testing graphing

dmann:


dmann:
Unfortunately, this solution will not work when integration tests require graphed persistence.

PR #756, opened 2016-10-10
Cherry picking the apidocs controller

dmann:


dmlond:
this breaks the build in the pull request. It might not work in production, which would create a potential merge conflict without the benefit of a serving the apiary in production.

PR #759, opened 2016-10-13
DDS-645 initiate project transfer

jonathanparrish:


app/models/project_transfer.rb
```diff
@@ -0,0 +1,26 @@
...
+  audited
+
+  belongs_to :project
+  belongs_to :from_user, class_name: 'User'
+
+  has_many :project_permissions, through: :project
+  has_many :project_transfer_users
+  has_many :to_users, through: :project_transfer_users
+
+  validates :project_id, presence: true
```
dmann: I think it would be better to validate against the association instead of the foreign key.

app/models/project_transfer.rb
```diff
@@ -0,0 +1,26 @@
...
+  validates :status, uniqueness: {
+      scope: [:project_id],
+      case_sensitive: false,
+      message: 'Pending transfer already exists'
+    }, if: :pending?
+  validates :from_user_id, presence: true
```
dmann: Same here regarding association vs foreign key validation.

app/models/project_transfer.rb
```diff
@@ -0,0 +1,26 @@
+class ProjectTransfer < ActiveRecord::Base
+  include RequestAudited
+  audited
+
+  belongs_to :project
+  belongs_to :from_user, class_name: 'User'
+
+  has_many :project_permissions, through: :project
+  has_many :project_transfer_users
+  has_many :to_users, through: :project_transfer_users
```
dmann, line 10: We may need to add a validation on either to_users or project_transfer_users to ensure that they're not empty.

spec/requests/project_transfer_api_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+
+  describe 'Project Transfer Collection' do
+    let(:url) { "/api/v1/projects/#{project_id}/transfers" }
+    let(:project_id) { project.id }
+
+    describe 'POST' do
+      subject { post(url, payload.to_json, headers) }
+      let(:called_action) { 'POST' }
+      let(:payload) {{
+        to_users: [
```
dmann, line 30: We should test that a validation error is thrown when to_users is empty. This may end up throwing a Grape validation error instead of an ActiveRecord validation error. Either is fine, though I'm not sure if the 'a validated resource' shared example will work well with the former.

dmann (changes requested):
Looks good over all. Just need a few validation tweaks and a test to account for bad input.

dmann (approved):

Issue #758, opened 2016-10-12
`GET/meta/{object_kind}/{object_id}{?meta_template_name}`  


caseychoiniere:
`GET /api/v1/meta/{object_kind}/{object_id}` endpoint to view the object metadata for a file, it doesn't matter what file ID I use, it always returns the same object metadata. This happens even when I use a file id from a file that has never had any object metadata added. Is this supposed to behave like this?

According to the docs the endpoint should be `GET/meta/{object_kind}/{object_id}{?meta_template_name}` with an optional param allowing for finding a template instance by name. But this param is missing from the API explorer. Still seems like it shouldn't be returning anything if the `file_id` provided has never had any metadata objects associated with it

PR #761, opened 2016-10-13
DDS-661 view all object metadata

dmann:


dmann:
This should address issue #758

dmlond (approved):
looks good to me

Issue #757, opened 2016-10-11
`is_deleted` property in this node is a string rather than a boolean


caseychoiniere:
@dmlond When a user doesn't have visibility to a provenance node (file_version), a partial `properties` resource is returned. But the `is_deleted` property in this node is a `string` rather than a boolean like the other provenance nodes. 

``` javascript
{
        "id": "a0bc22ce-4d19-4d04-af77-bca4e30902fd",
        "labels": [
          "FileVersion"
        ],
        "properties": {
          "id": "a0bc22ce-4d19-4d04-af77-bca4e30902fd",
          "kind": "dds-file-version",
          "is_deleted": "false"
        }
      },
```

 **versus**

``` javascript

"properties": {
          "kind": "dds-activity",
          "id": "61edc0b3-9fe2-4d64-9487-8853941a3ec6",
          "name": "test fileB non visible nodes",
          "description": "",
          "started_on": "2016-10-07T16:26:58.180Z",
          "ended_on": null,
          "is_deleted": false,
          "audit": {
            "created_on": "2016-10-07T16:26:58.326Z",
            "created_by": {
              "id": "e8aeeeb8-0071-4308-8bfa-e24d08094c1d",
              "username": "cc487",
              "full_name": "Casey Choiniere"
            },
            "last_updated_on": null,
            "last_updated_by": null,
            "deleted_on": null,
            "deleted_by": null
          }
```

which is causing inconsistencies in how the client handles the response.

PR #762, opened 2016-10-13
Dds 659 is deleted string

dmlond:


dmlond:
This addresses Issue #757

PR #772, opened 2016-10-20
Dds 665 move container creation

dmlond:
moves creation of project container in swift from POST /uploads to POST /projects
uses ActiveJob framework with inline execution of job to prepare us to use a queuing system to do this work in the background in the future


dmann (approved):
:P

PR #774, opened 2016-10-20
DDS-646 list project transfers

jonathanparrish:


app/api/dds/v1/project_transfers_api.rb
```diff
@@ -36,14 +36,26 @@ class ProjectTransfersAPI < Grape::API
...
           validation_error!(project_transfer)
         end
+      end
 
+      desc 'List project transfers' do
+        detail 'list project transfers'
+        named 'list project transfers'
+        failure [
+          [200, 'Success'],
+          [401, 'Unauthorized']
```
dmlond, line 51: since you are running project.find(id) you should include 

[404, 'Project does not exist'] 

to the list of failures.
One way to standardize this, anytime your spec states that an endpoint behaves like 'an identified resource' the failure codes should include a 404 entry with a description.

dmlond (approved):

PR #818, opened 2016-11-11
Dds 643 oit oauth

dmlond:
implemented ability to create an openid authentication service object to interact with duke oit openid OAUTH implementation


app/api/dds/v1/users_api.rb
```diff
@@ -13,59 +13,30 @@ class UsersAPI < Grape::API
...
-            end
-            ApiToken.new(user: authorized_user.user, user_authentication_service: authorized_user)
-          else
-            error!({
-              error: 401,
-              reason: 'invalid access_token',
-              suggestion: 'authenticaton service not recognized'
-            }, 401)
-          end
+        raise InvalidAccessTokenException.new unless token_info_params[:access_token]
```
dmann: I'm curious if this is needed. My instinct is that an exception like this would be thrown by the auth service model when get_user_for_access_token in called. Will the tests continue to pass if line 28 is removed?

app/models/duke_authentication_service.rb
```diff
@@ -0,0 +1,34 @@
+class DukeAuthenticationService < AuthenticationService
+  def get_user_for_access_token(encoded_access_token)
+    begin
+      access_token = JWT.decode(encoded_access_token, Rails.application.secrets.secret_key_base)[0]
+      user_authentication_service = user_authentication_services.where(uid: access_token['uid']).first
```
dmann: Does ActiveRecord have a find_by_uid that could be used instead of the where(uid: foo).first?

app/models/duke_authentication_service.rb
```diff
@@ -0,0 +1,34 @@
...
+      if user_authentication_service
+        user = user_authentication_service.user
+        user.current_user_authenticaiton_service = user_authentication_service
+        user
+      else
+        user = User.where(username: access_token['uid']).take
```
dmann: Similar to above, is there a User.find_by_username method?

lib/tasks/auth_service.rake
```diff
@@ -1,20 +1,85 @@
...
+
+namespace :auth_service do
+  desc "transfer default authentication_service from ENV[FROM_AUTH_SERVICE_ID] to ENV[TO_AUTH_SERVICE_ID]"
+  task transfer_default: :environment do
+    raise 'please set ENV[FROM_AUTH_SERVICE_ID] and ENV[TO_AUTH_SERVICE_ID]' unless ENV['FROM_AUTH_SERVICE_ID'] && ENV['TO_AUTH_SERVICE_ID']
+    from_auth_service = AuthenticationService.where(service_id: ENV['FROM_AUTH_SERVICE_ID']).take!
+    raise '#{from_auth_service.service_id} is not default' unless from_auth_service.is_default?
+
+    to_auth_service = AuthenticationService.where(service_id: ENV['TO_AUTH_SERVICE_ID']).take!
+    from_auth_service.update(is_default: false)
```
dmann: This is probably pretty low risk, but these updates should probably be in a transaction, with a failure rolling it all back and spitting out an error.

app/models/openid_authentication_service.rb
```diff
@@ -0,0 +1,42 @@
+class OpenidAuthenticationService < AuthenticationService
+  validates :client_id, presence: true
+  validates :client_secret, presence: true
+
+  def get_user_for_access_token(token)
+    user_info = get_userinfo(token)
+    uid = user_info['dukeNetID']
```
dmann, line 8: Lets store these user_info keys in the database!

spec/lib/tasks/auth_service_rake_spec.rb
```diff
@@ -0,0 +1,154 @@
+require 'rails_helper'
+
```
dmann, line 6: Since silence_stream is used later on in this file, add the following here for Rails 5 compatibility:

```
if Rails.version > '5.0.0'              
  require 'active_support/testing/stream'  
  include ActiveSupport::Testing::Stream
end
```

spec/support/authentication_service_shared_examples.rb
```diff
@@ -0,0 +1,452 @@
...
+
+shared_examples 'an authentication request endpoint' do
+  # requires the following be let
+  #   first_time_user
+  #   first_time_user_access_token
+  #   existing_user
+  #   existing_user_access_token
+  #   invalid_access_token
+
+  it_behaves_like 'a GET request' do
```
dmann, line 131: I envisioned the 'a GET request' shared example as a top level call in the main request spec files that wraps calls to more functional shared examples. I really should have documented it better.  :(

spec/support/authentication_service_shared_examples.rb
```diff
@@ -0,0 +1,452 @@
...
+          end
+          expect(error_response['error']).to eq(400)
+          expect(error_response['reason']).to eq('no access_token')
+          expect(error_response['suggestion']).to eq('you might need to login through an authentication service')
+        end
+      end
+
+      context 'nil access_token' do
+        let(:payload) {
+          if authentication_service.is_default?
```
dmann: An `if` statement in a test is a code smell. Let's look through this together and see what we can break off into its own shared example and refactor using specific contexts for default and non-default auth services.

dmann (changes requested):
I think it looks good overall. There are a couple things that should be added (like the Rails 5 compatibility bit) before merging this in to the develop branch. If time permits, it would be good for us to refactor users_api_spec and authentication_service_shared_examples.

app/api/dds/v1/users_api.rb
dmlond: refactored into the model

app/models/duke_authentication_service.rb
dmlond: yes

app/models/duke_authentication_service.rb
dmlond: yes

lib/tasks/auth_service.rake
dmlond: done

spec/lib/tasks/auth_service_rake_spec.rb
dmlond: done

spec/support/authentication_service_shared_examples.rb
dmlond: changed

dmlond: refactored to use service_id in payload

dmlond: refactored

dmann (approved):

PR #845, opened 2016-12-02
OAuth update and bug fixes

dmann:
- Support for multiple authentication providers
- Implementing pagination for endpoints listing `children` and `projects`
- Many bug fixes, mostly for metadata endpoints
- Parallelization of circle tests


dmlond (approved):
this is a general promotion of develop to ua_test according to our release policy that develop should always be ready to promote.

PR #846, opened 2016-12-02
DDS-651-view all project_transfers endpoint

jonathanparrish:


.gitignore
```diff
@@ -25,3 +25,5 @@ secrets.*.env
 docker/circle/ruby_2.2.2.docker.tgz
 docker/circle/dredd.docker.tgz
 docker-compose.override.yml
+
+coverage
```
dmann: I think you meant for this .gitignore change to be part of your other pull request.

dmann (changes requested):
Just the one thing with the gitignore. Maybe you can roll back the specific commit for that change? Otherwise, the rest of it looks good.

dmann (approved):

PR #847, opened 2016-12-05
Added simplecov to project

jonathanparrish:
Here is the simplecov install.

- Run your tests locally and then go to the coverage folder.  

- Next open the index.html file in your browser. 

Simplecov creates an interactive folder that separates the tests and their percentage of coverage by categories.  You can click on the file and it will pull up the code and highlight in red the code that is still in need of testing.


dmann (changes requested):
It looks like there could be some configuration done to allow the simplecov results to show up in circle: 
https://circleci.com/docs/code-coverage/#ruby

CircleCI is also running our tests in parallel, which means that simplecov needs to be configured to work in parallel. There is some information in the simplecov README.md on this. Unfortunately, it's not super clear how to do this, so you'll have to fiddle with it until it works as expected.

PR #853, opened 2016-12-08
DDS-738 serialize project creator parent

dmlond:


app/models/folder.rb
```diff
@@ -33,7 +33,7 @@ def is_deleted=(val)
   end
 
   def as_indexed_json(options={})
-    FolderSearchDocumentSerializer.new(self).as_json
+    Search::FolderSerializer.new(self).as_json
```
dmann, line 36: Namespacing FTW! 👍

app/models/data_file.rb
```diff
@@ -83,6 +83,27 @@ def as_indexed_json(options={})
...
+        indexes :username, type: "string"
+        indexes :first_name, type: "string"
+        indexes :last_name, type: "string"
+        indexes :email, type: "string"
+      end
     end
   end
+
+  def creator
+    current_file_version.audits.find_by(action: "create").user
```
dmann: Might want to put a guard clause around this in case current_file_version, .audits, or .find_by(action: "create") return nil.

app/models/folder.rb
```diff
@@ -43,6 +43,24 @@ def as_indexed_json(options={})
...
+        indexes :first_name, type: "string"
+        indexes :last_name, type: "string"
+        indexes :email, type: "string"
+      end
+
     end
   end
+
+  def creator
+    audits.find_by(action: "create").user
```
dmann: A guard clause would be good incase audits.find_by returns nil. Though, maybe it'll throw a Record Not Found exception, which could be better/worse?  :\

app/serializers/search/data_file_serializer.rb
```diff
@@ -0,0 +1,27 @@
...
+  has_many :tags, serializer: Search::TagSummarySerializer
+  has_one :project, serializer: Search::ProjectSummarySerializer
+  has_one :parent, serializer: Search::FolderSummarySerializer
+  has_one :creator, serializer: Search::UserSummarySerializer
+  
+  def is_deleted
+    object.is_deleted?
+  end
+
+  def meta
```
dmann, line 19: Could the `meta` method possibly work as a has_many? Not a criticism, I'm honestly curious.

circle.yml
```diff
@@ -46,6 +46,9 @@ deployment:
     - heroku run rake --exit-code db:seed --app 'dukeds-dev'
     - heroku run --exit-code rake db:data:migrate --app 'dukeds-dev'
     - heroku run --exit-code rake auth_service:set_default --app 'dukeds-dev'
+    - "[[ -z ${RECREATE_SEARCH_MAPPINGS} ]] || heroku run --exit-code rake elasticsearch:index:drop --app 'dukeds-dev'"
+    - "[[ -z ${RECREATE_SEARCH_MAPPINGS} ]] || heroku run --exit-code rake elasticsearch:index:create  --app 'dukeds-dev'"
+    - "[[ -z ${RECREATE_SEARCH_MAPPINGS} ]] || heroku run --exit-code rake elasticsearch:index:index_documents  --app 'dukeds-dev'"
```
dmann, line 51: Future us should create a single rake task to wrap the 3 elasticsearch tasks needed for recreating search mappings.

spec/models/data_file_spec.rb
```diff
@@ -245,48 +245,75 @@
...
+    subject {
+      Audited.audit_class.as_user(creator) do
+        FactoryGirl.create(:data_file)
+      end
+    }
+    it { is_expected.to respond_to :creator }
+    it {
+      expect(subject.current_file_version).not_to be_nil
+      expect(subject.current_file_version.audits).not_to be_empty
+      expect(subject.current_file_version.audits.find_by(action: 'create')).not_to be_nil
```
dmann: These 3 expectations (lines 257-9) could be broken out into their own contexts, which will ultimately test the guard clause in your creator method.

spec/models/folder_spec.rb
```diff
@@ -141,39 +141,55 @@
     end
   end
 
-  describe 'elasticsearch' do
-    it { expect(described_class).to include(Elasticsearch::Model) }
-
-    # TODO, when we move to asynchronous indexing, remove this and replace with
-    # a test to ensure that jobs are created on create, update, delete
-    it { expect(described_class).to include(Elasticsearch::Model::Callbacks) }
+  describe 'creator' do
```
dmann: Change to `describe '#creator' do`, we should try to follow the naming convention at http://betterspecs.org/#describe
So, `#` for instance methods and `::` for class methods. (I'm moving away from using the `.` as I've found that it's not explicit enough)

spec/models/folder_spec.rb
```diff
@@ -141,39 +141,55 @@
...
+    }
 
-    describe 'as_indexed_json' do
-      # let!(:tag) { FactoryGirl.create(:tag, taggable: child_folder) }
-      it { is_expected.to respond_to 'as_indexed_json' }
-      it { expect(subject.as_indexed_json).to eq(FolderSearchDocumentSerializer.new(subject).as_json) }
-    end
+    it { is_expected.to respond_to :creator }
+    it {
+      expect(subject.audits.find_by(action: 'create')).not_to be_nil
```
dmann: See the above comment regarding contexts and guard clauses.

dmann (changes requested):
Some changes and a few questions. Good work over all!

app/serializers/search/data_file_serializer.rb
dmlond: I dont know? It certainly doesnt lend itself to it. It wants a meta hash, with a series of key (meta_template.template.name): Hash with series of key (meta_property.property.key): meta_property.value entries. These do not align nicely with the underlying object structures

circle.yml
dmlond: good idea

spec/models/folder_spec.rb
dmlond: thanks. will do this

app/models/folder.rb
dmlond: 👍

app/models/data_file.rb
dmlond: 👍

spec/models/data_file_spec.rb
dmlond: 👍

spec/models/folder_spec.rb
dmlond: 👍

spec/models/folder_spec.rb
dmlond: 👍

dmann (approved):

PR #855, opened 2016-12-12
Move dredd tests to dds development.

benneely:
Previously dredd tests were run against a dockerized version of dds. Now, they will hit the development server after development deploy from circle.


circle.yml
```diff
@@ -80,3 +80,5 @@ deployment:
     - "[[ -z ${RECREATE_SEARCH_MAPPINGS} ]] || heroku run --exit-code rake elasticsearch:index:drop --app 'dukeds'"
     - "[[ -z ${RECREATE_SEARCH_MAPPINGS} ]] || heroku run --exit-code rake elasticsearch:index:create  --app 'dukeds'"
     - "[[ -z ${RECREATE_SEARCH_MAPPINGS} ]] || heroku run --exit-code rake elasticsearch:index:index_documents  --app 'dukeds'"
+    - echo "apiToken="`heroku run rake api_test_user:create -a dukeds-dev | tail -1 | sed 's/\*\*.*\[NewRelic\].*//'` >> dredd.env
```
dmlond, line 83: could we wrap lines 82 and 83 into circle/run_dredd.sh. Also, do we want to clean up everything at the end, or let test artifacts build up in the development db/swift/neo4j/elasticsearch over time?

PR #856, opened 2016-12-13
DDS-755 rebuild index with env toggle

dmlond:


dmann (approved):
Looks great, good job! 👍

PR #857, opened 2016-12-13
DDS-728 concurrent save without dups

dmann:
Preventing creation of duplicate MetaTemplate with a row level lock on the associated Template object.


dmlond (approved):
nice use of Thread

dmlond:
:( circle integration failed for this PR. the build passed, but the PR status still shows it is running

PR #871, opened 2017-01-03
fixed bug in the authservice service

dmlond:
- needs to run auth_service:duke_create instead of authentication_service:create
  - auth_service.local.env needed other ENV to meet new requirements


dmlond:
this also replaces all the
```
volumes_from:
  - server
```

with
```
  volumes:
    - .:/var/www/app
```
which prevents docker-compose from launching a server every time rspec, rake, etc. are run

dmann (commented):
I think the auth_service env vars and command update are a good add. 

Changing from volumes_from to volumes seems like a step backward though. I understand the intent, but I think it would make more sense to setup a service definition like 'server_volumes' or 'app_volumes' that has the `.:/var/www/app` volume, and then update server and the rest of the services to use volumes_from this new service. Does that make sense?

dmann (approved):
Looks great! 👍

PR #873, opened 2017-01-10
DDS-761 harmonize get elastic serializers

dmlond:
everything that is returned about a DataFile or Folder from the endpoints is also serialized in the elasticsearch indices


dmann (approved):
Looks like you really DRYed out a good deal of the serializer code. Good job! 👍

PR #874, opened 2017-01-10
DDS-765 new mappings

dmlond:
This should have gone with DDS-761
The mappings needed to be updated


dmann (approved):
Looks good. I would like to DRY out this code in the future. There's a lot of repetition and I'm curious what could be inferred from other relations.

PR #877, opened 2017-01-12
DDS-774 keyword indexing

dmlond:
changes kind, id indices to strings that are not_analyzed, for keyword searches.
Also changes some things like size and version to integer.
hash.algorithm and hash.value are also strings not analyzed.


dmann (approved):
Looks good. 👍

Issue #878, opened 2017-01-17
Can not access custom response headers included in paginated results


caseychoiniere:
can't access the custom headers that are included in the response for results that are paginated.

the API docs say:
```To make it easy for clients to manage pagination, the following headers will always be included in the response:

X-Total: The total number of objects that would be returned by the query unpaginated
X-Total-Pages: The total number of pages given X-Total and X-Per-Page
X-Page: Current Page (passed in as query parameter)
X-Per-Page: per_page (passed in as query parameter)
X-Next-Page: Next page to be collected
X-Prev-Page: Previous page (will be empty on first page)
```

The only headers that are available are `cache-control` and `content-type`. Can you take a look at this Stack Overflow and read the accepted answer. I believe it's relevant to the issue I'm having and may require an additional change on your end?

http://stackoverflow.com/questions/38953864/how-to-get-headers-of-the-response-from-fetch

From Mozilla: https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS#Access-Control-Expose-Headers

Also here is a snippet from a Google developers page that also explains the restrictions on what headers can be viewed in a CORS response. 
```
a cors response restricts the headers you can view to `Cache-Control`, `Content-Language`, `Content-Type`, `Expires`, `Last-Modified`, and `Pragma`.****
```

caseychoiniere:
Thanks!

PR #879, opened 2017-01-17
Exposing the pagination headers via Rack::Cors

dmann:
Addresses Issue #878 

Since pagination is a core functionality and won't change from one deployment to another, I decided it was better to code the header values into the config as opposed to setting via ENV.


dmlond (approved):
👍

PR #880, opened 2017-01-17
DDS-754 swagger oit oauth

dmlond:
implements OIT Authentication in the swaggerui frontend
This will no longer be able to use the older Duke Data Service authentication


dmann (approved):
Looks good. 👍

PR #883, opened 2017-01-19
DDS-731 align kind relations

dmlond:
I removed many of the kind shared_example expectations
This requires us to explicitly define them in the models, either in the kind definition, or the 
ProvRelation definition. This is actually better, as it forces us to clearly state the expected kind that is reported in the api documentation.


dmann (approved):
👍

PR #894, opened 2017-01-26
DDS-784 log level configuration develop

dmann:


dmlond (approved):
We want to monitor dev closely if we set the log level to :debug. We had memory issues last time for reasons that I cannot recall.

PR #898, opened 2017-01-27
changing the DataFile upload.size mapping

dmlond:
from integer (2^^32-1) to long (2^^63-1)

https://github.com/karmi/retire/issues/474


dmann (approved):
🥇

PR #902, opened 2017-02-02
DDS-715 auth service affiliates

dmlond:
implements:
- get /auth_services/:id/affiliates
- get /auth_services/:id/affiliates/:uid
- post /auth_services/:id/affiliates/:uid/dds_user

adds concept of affiliate_search to auth_providers
auth_providers support affiliate_search if they belong_to an identity_provider

affiliate search endpoints return an exception if the auth_service does not support affiliate_search


dmlond:
I need to add rake tasks for identity_provider management. Hold off on the review

app/api/dds/v1/auth_providers_api.rb
```diff
@@ -28,6 +57,113 @@ class AuthProvidersAPI < Grape::API
...
+          authentication_service_id: params[:id],
+          uid: affiliate_params[:uid]
+        ).exists?
+          affiliate_exists_error!
+        else
+          auth_service = AuthenticationService.find(params[:id])
+          if auth_service.identity_provider.nil?
+            unsupported_affiliate_search_error!
+          else
+            auth_service.identity_provider.affiliates(affiliate_params[:full_name_contains])
```
dmann: I think we're going to have to figure out a way to paginate these results at some point.

dmann: Since this will be a breaking change, should we paginate before merging?

app/models/identity_provider.rb
```diff
@@ -0,0 +1,4 @@
+class IdentityProvider < ActiveRecord::Base
```
dmann, line 1: Since `IdentityProvider` is an interface, I think it would be good to add `affiliate` and `affiliates` methods that raise NotImplementedError, to ensure that child classes implement them.

app/models/ldap_identity_provider.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  def affiliates(full_name_contains=nil)
+    return [] unless full_name_contains && full_name_contains.length > 3
+    ldap_search(
+      Net::LDAP::Filter.construct("displayName=*#{full_name_contains}*")
+    )
+  end
+
+  def ldap_search(ldap_filter)
```
dmann: I think `ldap_search` should be private.

spec/factories/users.rb
```diff
@@ -1,7 +1,7 @@
 FactoryGirl.define do
   factory :user do
     etag { SecureRandom.hex }
-    sequence(:username) { |n| "#{Faker::Internet.user_name}_#{n}" }
+    sequence(:username) { |n| "#{Faker::Name.first_name}_#{n}" }
```
dmann: `Faker::Name.first_name` includes names with apostrophes, like D'Angelo, which can lead to transient test failures.  :(

spec/models/ldap_identity_provider_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+  subject { auth_provider.identity_provider }
+  let(:auth_provider) { FactoryGirl.create(:openid_authentication_service, :with_ldap_identity_provider) }
+  let(:test_user) { FactoryGirl.attributes_for(:user) }
+
+  describe 'validations' do
+    it { is_expected.to validate_presence_of :ldap_base }
+  end
+
+  describe 'affiliate' do
+    it { is_expected.to respond_to(:affiliate) }
```
dmann: You can get rid of the following context by replacing the above with:
```
it { is_expected.to respond_to(:affiliate).with(1).argument }
```

dmann: https://www.relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/respond-to-matcher

spec/models/ldap_identity_provider_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+
+  subject { auth_provider.identity_provider }
+  let(:auth_provider) { FactoryGirl.create(:openid_authentication_service, :with_ldap_identity_provider) }
+  let(:test_user) { FactoryGirl.attributes_for(:user) }
+
+  describe 'validations' do
+    it { is_expected.to validate_presence_of :ldap_base }
+  end
+
+  describe 'affiliate' do
```
dmann: Nit picky, but please prepend method names with `#` for instance and `::` for class methods.

spec/models/ldap_identity_provider_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+      it {
+        is_expected.to be
+        is_expected.to be_a User
+        is_expected.not_to be_persisted
+        expect(subject.display_name).to eq test_user[:display_name]
+      }
+    end
+  end
+
+  describe 'affiliates' do
```
dmann: Please prepend method name with `#`

spec/support/ldap_shared_examples.rb
```diff
@@ -0,0 +1,91 @@
...
+      expected_entry[:uid] = test_user[:username]
+      expected_entry[:givenName] = test_user[:first_name]
+      expected_entry[:sn] = test_user[:last_name]
+      expected_entry[:mail] = test_user[:email]
+      expected_entry[:displayName] = test_user[:display_name]
+      expected_entry
+    }
+  }
+  before do
+    allow_any_instance_of(Net::LDAP).to receive(:search).and_return(expected_entries)
```
dmann, line 15: Cool! I'm sure there are other places in our code that could use this technique.

dmann (changes requested):
Looks good for the most part. Just asking for a few minor changes.

app/api/dds/v1/auth_providers_api.rb
dmlond: go ahead and merge. We will need to work through a cycle of customer feedback, Matt specifying API design for pagination, developer spiking LDAP pagination and implementation as a new feature.

dmlond:
all but pagination are addressed

app/api/dds/v1/auth_providers_api.rb
dmann: @Donmurry what do you think?

Donmurry: Hmm, I think it might be best to add pagination in now, if possible.  If spike of LDAP pagination is going to take weeks vs days, then i say we can skip it.  But if we can incorporate it now, I think thats better than not.  We can weigh the cost of development now vs cost later.

dmann: If the decision is to skip pagination, it would be good to inform the client developers of this decision and the reasoning behind it. I think they would appreciate a warning that, if they submit a query with 3 characters, they're likely to get a firehose of data OR at request timeout. That way when both of those outcomes eventually happen, hopefully, they won't make an "issue" out of it...

dmlond: https://github.com/ruby-ldap/ruby-net-ldap/blob/e4c46a223a19feda78393a793711353aa1febdcd/lib/net/ldap/connection.rb

I really dont think the ruby net-ldap search interface allows you access to pagination. They paginate the results under the hood, but they repeatedly search until all results are returned before returning to the caller with all the results.  They do allow you to send a maximum size, and time value (this might be useful to prevent timeouts), but these just end up minimizing each call to get the next page's results, while still continuously getting each pages results until they are all returned.

If we want to support pagination, we will have to do it wastefully, e.g. get all results, skip to N, add until M, break. This would satisfy clients requirement if they end up needing pagination, but would not prevent server timeouts of memory explosions.

dmann: I feel strongly that this endpoint needs to limit the number of returned results. The pagination default is 100, with a max of 1000. Ideally, the client would have the ability change this limit via the `per_page` parameter.

While it would be nice to do a full implementation of pagination, I don't believe that is necessary. That is, I don't think that we need to allow clients to use the `page` parameter. In fact, if `page` is provided, I suggest returning a "Not Implemented" or "Not supported by Identity Provider" error of some kind.

This is a technical limitation of ldap and that shouldn't dictate how our interface works. Let's say that you initially setup the affiliates endpoints with an identity provider that supported pagination and you were easily able to add support for `per_page` and `page`. Feature released, done. Now, you have to add a second identity provider that does not support pagination. What do you do? Remove pagination from the endpoint? Probably not. This is the way that I'm approaching this problem and hopefully provides context to my suggestions above.

dmann (approved):

PR #911, opened 2017-02-10
DDS-809 fix request timeout develop

dmann:
The puma worker_timeout configuration only works in cluster mode, which I don't think our deployment uses:

http://www.rubydoc.info/github/puma/puma/Puma%2FConfiguration%2FDSL%3Aworker_timeout

We should use the Rack::Timeout gem as described here:

https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server#timeout

https://github.com/heroku/rack-timeout/tree/v0.4.2


dmlond (dismissed):
we all need to remember to docker-compose `build server`

dmlond (approved):
👍

dmann:
WTF Github!

PR #913, opened 2017-02-10
DDS-809 fix request timeout prod

dmann:
The puma worker_timeout configuration only works in cluster mode, which I don't think our deployment uses:

http://www.rubydoc.info/github/puma/puma/Puma%2FConfiguration%2FDSL%3Aworker_timeout

We should use the Rack::Timeout gem as described here:

https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server#timeout

https://github.com/heroku/rack-timeout/tree/v0.4.2

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -18,6 +18,7 @@ gem 'puma'
 gem 'rack', '1.6.4' # Remove in rails 5
 gem 'rack-cors', :require => 'rack/cors'
 gem 'grape-middleware-lograge'
+gem "rack-timeout"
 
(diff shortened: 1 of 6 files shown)


dmann:
Please wait to merge this. Rack::Timeout works, but returns a 500 error instead of a 503. I would like to create a test and ensure that a 503 with a reasonable message is returned.

dmann:
This will be applied when v1.1.2 is merged to production via ua_test

PR #915, opened 2017-02-14
[Please don't merge] - Add S3 as a storage provider

colakong:
Hi DDS devs,

I've abstracted out Swift and S3 as storage providers, using Ruby mixins. Operations I've tested for Swift all work. S3 operations like creating a bucket, object work, but the steps after object creation fail when using S3 provider (`get_object_metadata`). It looks like other models (like `upload.rb`) assume a Swift-like process for uploading chunks, which causes confusion when used blindly with S3 provider (code ends up saving chunks as objects instead of an object-of-chunks, and then tries to get metadata for an object-of-chunks that doesn't exist)

In order to get it fully working I think I'd also need to abstract out `upload.rb` into Swift-style and S3 styles of uploads too. Maybe for S3 we'd create and keep track of an S3 multi-part upload object, and use that with each call to `/uploads/{id}/chunks`.

I wanted to create the PR so that you could see what I've done so far, and maybe let me know if this is the right approach. This is my first time using Ruby and Ruby on Rails :blush:


dmlond:
Hi,
This is really cool. Thanks for working this out. We plan on adding S3 support in our roadmap, but we are not quite ready to do that yet. I would really like to put this code in a separate branch so that we can have it ready when we do go forward. I am going to create an s3_storage_provider branch, and close this PR. Can you please submit a new PR from your s3_storage_provider branch to ours. I would really appreciate it.

dmlond:
also, we are in the process of upgrading our versioning system.  I created our s3_storage_provider branch off our latest stable version (v1.1.1). You might need to change your s3_storage_provider branch to reflect what is in v1.1.1. I am not sure when you forked from our develop, but it will not likely be that different.

PR #918, opened 2017-02-14
Add S3 storage provider

colakong:
Hi @dmlond, DDS devs,

Here's a PR for adding S3 as a storage provider to DDS. It's based on the v1.1.1 tag, and leaves out some of the dev-environment stuff I had in [the previous PR](https://github.com/Duke-Translational-Bioinformatics/duke-data-service/pull/915).

Like I mentioned in the previous PR, other models (like `upload.rb`) would still need to be updated for it to work properly.


dmlond:
Thanks so much.

PR #927, opened 2017-02-16
[DEVELOP] Using Batched Elasticsearch Indexing

dmlond:


dmann (approved):
...batches of 500 or FEWER!!!

PR #936, opened 2017-02-20
Bug Fix: random auth_provider test failures

dmann:
AuthProvider#affiliates allows full_name_contains to be 3 characters

### DIFF
--- a/app/models/ldap_identity_provider.rb
+++ b/app/models/ldap_identity_provider.rb
@@ -10,7 +10,7 @@ class LdapIdentityProvider < IdentityProvider
   end
 
   def affiliates(full_name_contains=nil)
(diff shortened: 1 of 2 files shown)


app/models/ldap_identity_provider.rb
```diff
@@ -10,7 +10,7 @@ def affiliate(uid)
   end
 
   def affiliates(full_name_contains=nil)
-    return [] unless full_name_contains && full_name_contains.length > 3
+    return [] unless full_name_contains && full_name_contains.length >= 3
```
dmlond, line 13: my bad

dmlond (approved):
thanks for fixing :1st_place_medal:

PR #941, opened 2017-02-21
DDS-810 remove portal from api repo

dmann:
Recompile Gemfile.lock


dmlond (approved):
I wonder if there is a local [git hook](http://githooks.com/) we could write to check that Gemfile.lock is changed if Gemfile is changed?

PR #942, opened 2017-02-21
Merge v1.2.3, v1.2.4, and v1.2.5 to ua_test

dmann:
- AuthProvider affiliates bug fix
- Removing the portal from the api repo
- Redirect root url to `/apidocs`


config/routes.rb
```diff
@@ -1,5 +1,4 @@
 Rails.application.routes.draw do
   mount DDS::Base, at: '/'
   get "apiexplorer" => "swaggerui#index"
-  root to: redirect("/portal", status: 302)
```
dmlond, line 4: we really need a root. Otherwise, if someone navigates to  https://apidev.dataservice.duke.edu, they get the default rails application

dmlond (changes requested):

config/routes.rb
dmann: I created a Bug issue in Jira. We can leave this PR open until the fix is applied to develop.

dmann: @dmlond where do you think the root url should redirect to? The apidocs or apiexplorer? I think I'm leaning more towards the former, since it contains more information about DDS.

dmlond (approved):
👍

PR #943, opened 2017-02-22
DDS-800_recursive_deletes

dmlond:


dmann (changes requested):
Change FolderDeletionJob to ChildDeletionJob and move perform later to manage_children. Pass in serialized parent object instead of folder_id.

dmlond:
I made the requested changes. Thanks for the suggestions.

spec/factories/uploads.rb
```diff
@@ -1,7 +1,7 @@
 FactoryGirl.define do
   factory :upload do
     project
-    name { Faker::Internet.slug }
+    name { "#{Faker::Internet.slug}#{rand(10**3)}" }
```
dmann: For uniqueness in a factory attribute, use a sequence:

``
sequence(:name) { |n| "#{Faker::Internet.slug(nil, '_')}_#{n}" }
``

Using rand will still create dups from time to time.

app/jobs/child_deletion_job.rb
```diff
@@ -0,0 +1,9 @@
+class ChildDeletionJob < ApplicationJob
+  queue_as self.name.underscore.gsub('_job','').to_sym
```
dmann: This is a little complex for what amounts to `queue_as :child_deletion`  :\

dmann (changes requested):
I really like what you've done with the ChildMinder module and the ChildDeletionJob. One minor change to the uploads factory to prevent possible test failures.

dmlond:
ready for review again

Procfile
```diff
@@ -1,2 +1,3 @@
 web: bundle exec puma -C config/puma.rb
-project_storage_init: rake workers:initialize_project_storage:run
+project_storage_init: bundle exec rake workers:initialize_project_storage:run
+folder_deletion: bundle exec rake workers:delete_folders:run
```
dmann: This should probably be renamed to `child_deletion` and `delete_children`

lib/tasks/workers.rake
```diff
@@ -10,4 +10,15 @@ namespace :workers do
...
     end
   end
+
+  namespace :delete_folders do
+    desc 'run a FolderDeletionJob'
+    task run: :environment do
+      silence_warnings do
+        Rails.application.eager_load! unless Rails.application.config.eager_load
+      end
+      workers = [ FolderDeletionJob.job_wrapper ]
```
dmann: I'm surprised that this isn't causing tests to fail, since FolderDeletionJob no longer exists...

spec/lib/tasks/workers_rake_spec.rb
```diff
@@ -6,21 +6,18 @@
...
+    let(:expected_job_class) { ProjectStorageProviderInitializationJob }
 
     it { expect(subject.prerequisites).to  include("environment") }
+    it_behaves_like 'a queued job worker', :expected_job_class
+  end
+
+  describe 'workers:delete_folders:run' do
+    include_context "rake"
+    let(:task_name) { "workers:delete_folders:run" }
+    let(:expected_job_class) { FolderDeletionJob }
```
dmann: Are the rake tests running in circle? Because this really should be failing...

dmann (changes requested):
Sorry that I didn't catch these in my last review. Lesson learned, review all files changed and not just the most recent changes...  :(

lib/tasks/workers.rake
dmlond: good catch. These dont run in circle.

dmlond:
ready for review again

dmann (approved):
👍

PR #949, opened 2017-03-02
DDS-813 set sneakers job handler

dmann:


dmlond (approved):
👍 let me know if I should add anything to the /api/v1/app/status healthcheck

dmann:
@dmlond The exchange and queue named `active_jobs-error` should exist and will be used by all of our active jobs when retries have failed. The Maxretries handler will also create a `-retry` exchange, `-retry` queue, and `-retry-requeue` exchange for each active job queue, prepending the queue's name to each. It's probably a good idea to have these tracked in `/app/status`.

dmlond:
created jira card for new healthchecks

PR #950, opened 2017-03-07
DDS-835 message log worker

dmann:
The MessageLogWorker takes messages off the 'message_log' queue and sends them to Rails.logger.info as a way of keeping a record of RabbitMQ messages, which will hopefully be useful for debugging purposes.

### DIFF
--- a/spec/workers/message_log_worker_spec.rb
+++ b/spec/workers/message_log_worker_spec.rb
@@ -14,5 +14,27 @@ RSpec.describe MessageLogWorker do
   it { is_expected.to respond_to(:work_with_params).with(3).arguments }
 
   describe '#work_with_params' do
+    let(:message) { Faker::Lorem.words(5) }
+    let(:delivery_info) { {exchange: Faker::Internet.slug, id: Faker::Number.digit} }
+    let(:metadata) { :baz }
+    let(:method_call) { subject.work_with_params(message, delivery_info, metadata) }
+    let(:ack) { subject.ack! }
+    it { expect(ack).not_to be_nil }
+    it { expect(method_call).to eq ack }
+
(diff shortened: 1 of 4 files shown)


spec/workers/message_log_worker_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    context 'calls Rails.logger' do
+      let(:log_message) {{
+        MESSAGE_LOG: {
+          message: message,
+          delivery_info: delivery_info,
+          metadata: metadata
+        }
+      }.to_json}
+      before do
+        allow(Rails.logger).to receive(:info)
```
dmlond: remove the before block entirely (lines 33-36)
change the it block at line 37 to
it {
  expect(Rails.logger).to receive(:info).with(log_message)
  expect{ method_call }.not_to raise_error
}
this will fail if the method_call does not call Rails.logger.info(log_message)

dmlond (changes requested):

spec/workers/message_log_worker_spec.rb
dmlond: if your method_call calls Rails.logger.info with any other messages, you will need to do
expect(Rails.logger).to receive(:info).with(log_message)
allow(Rails.logger).to receive(:error).with(any_args)
...

dmann: I tried it both ways as I was writing it. I found that using the before block made it more readable and allowed more flexibility when testing the arguments being passed. For instance, I can have 3 different regex tests, one for message, delivery_info, and metadata. But, since it's only a single test now, I'll reduce the context block to a single it test.  :P

dmlond (approved):
👍

PR #955, opened 2017-03-10
DDS-801 elasticsearch jobs

dmlond:


app/api/dds/v1/app_api.rb
```diff
@@ -55,33 +51,36 @@ class AppAPI < Grape::API
...
-            missing_exchanges = [
-              ApplicationJob.opts[:exchange],
+            [
+              Sneakers::CONFIG[:exchange],
+              Sneakers::CONFIG[:retry_error_exchange],
               ApplicationJob.distributor_exchange_name
-            ].reject do |expected_exchange|
-              ApplicationJob.conn.exchange_exists? expected_exchange
+            ].each do |expected_exchange|
+              unless ApplicationJob.conn.exchange_exists? expected_exchange
```
dmann, line 63: We need to change this to use something other than the (private) '.conn' method. A future PR will allow the usage of Sneakers::CONFIG[:connection] for this purpose.

lib/tasks/workers.rake
```diff
@@ -28,4 +28,15 @@ namespace :workers do
       Sneakers::Runner.new(workers).run
     end
   end
+
+  namespace :index_documents do
+    desc 'run an ElasticsearchIndexJob'
+    task run: :environment do
+      silence_warnings do
+        Rails.application.eager_load! unless Rails.application.config.eager_load
```
dmann, line 36: We may not need to call eager_load! here, since we're not using inflection to find the job classes. Removal can be explored in a future PR.

dmann (approved):
Added a couple comments as reminders of things we should take care of in the near future.

PR #959, opened 2017-03-16
better error handling in ldap search

dmlond:
gracefully trancates results when they would be too large for the supporting
ldap server.
No more than 500 results will ever be returned for a query.


dmann (approved):
👍

PR #962, opened 2017-03-17
Remove unneeded call to Rails.application.eager_load!

dmann:
This removes unneeded calls to `Rails.application.eager_load!` from the job rake tasks. The `eager_load!` calls were necessary when iterating over the descendants of a class (like `ApplicationJob.descendants`), but not when calling those descendants by name.

Unfortunately, the `/app/status` endpoint and specs use `ApplicationJob.descendants` to test for RabbitMQ queue presence, which requires eager loading to work properly. So, in order to get predictable test results, I enabled eager loading in the rails test environment config. Incidentally, the production environment is configured to eager load by default, so I think this will lead to fewer deployment surprises down the road.


dmlond (approved):
🔢

PR #964, opened 2017-03-24
DDS-855 job transactions

dmlond:


app/api/dds/v1/folders_api.rb
```diff
@@ -32,7 +32,10 @@ class FoldersAPI < Grape::API
           name: folder_params[:name]
         })
         authorize folder, :create?
-        if folder.save
+        Folder.transaction do
```
dmann: I don't think the transaction block is needed.

app/api/dds/v1/folders_api.rb
```diff
@@ -69,7 +72,9 @@ class FoldersAPI < Grape::API
         authenticate!
         folder = hide_logically_deleted Folder.find(params[:id])
         authorize folder, :destroy?
-        folder.update(is_deleted: true)
+        Folder.transaction do
```
dmann: I don't think the transaction block is needed.

app/api/dds/v1/projects_api.rb
```diff
@@ -27,13 +27,10 @@ class ProjectsAPI < Grape::API
...
           creator_id: current_user.id,
         })
-        storage_provider = StorageProvider.first
-        if project.save
-          project.set_project_admin
-          ProjectStorageProviderInitializationJob.perform_later(
-            storage_provider: storage_provider,
-            project: project
-          )
+        Project.transaction do
```
dmann: I don't think the transaction block is needed.

app/api/dds/v1/projects_api.rb
```diff
@@ -122,7 +119,9 @@ class ProjectsAPI < Grape::API
         authenticate!
         project = hide_logically_deleted Project.find(params[:id])
         authorize project, :destroy?
-        project.update(is_deleted: true)
+        Project.transaction do
```
dmann: I don't think the transaction block is needed.

app/jobs/application_job.rb
```diff
@@ -1,4 +1,6 @@
 class ApplicationJob < ActiveJob::Base
+  include JobTracking
```
dmann, line 2: I like this. I'd like to do something similar to encapsulate the Sneakers specific logic in this class.

app/models/concerns/job_tracking.rb
```diff
@@ -0,0 +1,39 @@
...
+      raise ArgumentError.new("object is not job_transactionable") unless transactionable.class.include?(JobTransactionable)
+      JobTransaction.create(
+        transactionable: transactionable,
+        request_id: transactionable.current_transaction.request_id,
+        key: self.transaction_key,
+        state: 'initialized'
+      )
+    end
+
+    def self.start_job(initial_transaction)
```
dmann: `::start_job` and `::complete_job` are very repeaty. A `::record_transaction` method (named whatever makes sense) could be used to standardize the creation of JobTransactions. Though it seems to be slightly different, `::initialize_job` may also be able to use this new method.

dmann: Just scrolled down. Maybe `JobTransactionable#create_transaction` could be used to handle this refactor?

spec/jobs/project_storage_provider_initialization_job_spec.rb
```diff
@@ -5,35 +5,35 @@
...
-
-    it 'should require a storage_provider argument' do
-      expect {
-        described_class.perform_now
-      }.to raise_error(ArgumentError)
-      expect {
-        described_class.perform_now(project: project)
-      }.to raise_error(ArgumentError)
-    end
+  it {
```
dmann, line 15: These `expect { ... }.to raise_error` calls should be in their own `it {}` example blocks.

spec/models/project_spec.rb
```diff
@@ -56,14 +57,16 @@
     end
   end
 
-  describe 'set_project_admin' do
+  describe '.set_project_admin' do
```
dmann: Should be '#set_project_admin' since it's an instance method.

spec/requests/projects_api_spec.rb
```diff
@@ -55,9 +55,7 @@
           let(:resource) { project_stub }
           it 'should set creator to current_user and make them a project_admin, and queue an ActiveJob' do
             expect {
-              expect {
-                is_expected.to eq(201)
-              }.to have_enqueued_job(ProjectStorageProviderInitializationJob)
```
dmann, line 60: Did you remove the `have_enqueued_job` test because the job is being enqueued in an `after_commit` callback? If so, it may be worth looking into the `test_after_commit` gem (until we upgrade to Rails 5), which is mentioned at the end of this article: 
http://www.justinweiss.com/articles/a-couple-callback-gotchas-and-a-rails-5-fix/

spec/requests/projects_api_spec.rb
```diff
@@ -118,17 +116,20 @@
...
-          ActiveJob::Base.queue_adapter = :inline
-        end
-
-        let!(:storage_provider) { FactoryGirl.create(:storage_provider, :swift) }
-        it_behaves_like 'a creatable resource' do
-          let(:resource) { project_stub }
-          it_behaves_like 'a storage_provider backed resource'
-        end
-      end
+      # this test does not work because the after_commit on: :create
```
dmann, line 119: See above comment on `after_commit`

dmann (changes requested):
I'm not quite finished reviewing, but I wanted to submit what I have thus far. I made it through spec/requests/projects_api_spec.rb

spec/support/concerns/job_transactionable_shared_examples.rb
```diff
@@ -0,0 +1,58 @@
...
+      .before(:create)
+    is_expected.to callback(:root_update_transaction)
+      .before(:update)
+  }
+
+  describe '.create_transaction' do
+    let(:transaction_state) {'testing'}
+
+    context 'with nil current_transaction' do
+      it {
```
dmann, line 19: It would be good to test that `#current_transaction` is nil before calling `#create_transaction`

spec/support/concerns/job_transactionable_shared_examples.rb
```diff
@@ -0,0 +1,58 @@
...
+        expect(subject.current_transaction.state).to eq(transaction_state)
+      }
+    end
+
+    context 'with preset current_transaction' do
+      let(:existing_current_transaction)  {
+        subject.create_transaction 'pretest'
+      }
+
+      it {
```
dmann, line 34: It would be good to test the contents of `#current_transaction`. Also, `existing_current_transaction` isn't called, so I'm guessing that the current transaction is actually nil.  :\

spec/support/job_shared_examples.rb
```diff
@@ -7,31 +7,41 @@
...
+  it {
+    expect{described_class.perform_now}.to raise_error(ArgumentError)
+  }
   before {
     ActiveJob::Base.queue_adapter = :test
   }
   include_context 'with job runner', described_class
 
   context 'update' do
+    let(:existing_container) { FactoryGirl.create(container_sym) }
```
dmann, line 19: Good call moving this to a let var! 👍

spec/support/job_shared_examples.rb
```diff
@@ -7,31 +7,41 @@
...
+      include_context 'tracking job', :job_transaction
       it {
-        new_container = FactoryGirl.create(container_sym)
+        expect(new_container).to be_persisted
         mocked_proxy = double()
         expect(mocked_proxy).to receive(:index_document)
         expect(new_container).to receive(:__elasticsearch__).and_return(mocked_proxy)
-        ElasticsearchIndexJob.perform_now(new_container)
+        ElasticsearchIndexJob.perform_now(job_transaction, new_container)
+        job_transaction.reload
```
dmann: What is the call to `#reload` doing here? I'm confused because there aren't any expectations after it.

dmann (changes requested):
Just finished going through the specs, just a couple more questions and comments.

app/models/concerns/job_tracking.rb
dmlond: JobTransactionable applies to Models, JobTracking applies to Jobs context for create_transaction (transaction is only built, notcreated, and self is a Folder, Project, File instance) is different than initialize/start/complete/register_job (transaction is created, and self is a Job class)

spec/requests/projects_api_spec.rb
dmlond: ok. will look into it.

dmann (approved):
👍

PR #968, opened 2017-04-01
DDS-864 refactor background job workers

dmann:
Moving the logic for running jobs into a separate class "that we own", which will make mocking tests much more sane. This new class, `JobsRunner`, will keep track of all our jobs in a `workers_registry` hash, which is used in dynamically creating rake tasks for each job as well as in its own `::all` method. Currently `JobsRunner` is tied to `Sneakers::Runner`, but this could easily be teased out into a module in order to support different queue adapters.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,4 +1,5 @@
 web: bundle exec puma -C config/puma.rb
+all_workers: bundle exec rake workers:all:run
 message_log_worker: bundle exec rake workers:message_logger:run
 project_storage_init_job: bundle exec rake workers:initialize_project_storage:run
 child_deletion_job: bundle exec rake workers:delete_children:run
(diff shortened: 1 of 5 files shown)


dmlond:
you probably need to add something like the following to config/application.rb
```
  config.paths.add File.join('app', 'services'), glob: File.join('**', '*.rb')
  config.autoload_paths += Dir[Rails.root.join('app', 'services', '*')]
```

dmlond (approved):
this looks great. 👍

Procfile
```diff
@@ -1,4 +1,5 @@
 web: bundle exec puma -C config/puma.rb
+all_workers: bundle exec rake workers:all:run
 message_log_worker: bundle exec rake workers:message_logger:run
```
dmlond, line 3: do the other workers still work? if not, we should remove these lines from the Procfile

dmlond (approved):
looks good 👍

PR #969, opened 2017-04-03
DDS-866-remove_vhost

dmlond:
removing vhost bunny configuration, as it sets it automatically from the ENV['CLOUDAMQP_URL']


dmann (approved):
💯

PR #970, opened 2017-04-04
Bundle update gems

dmann:
Updating ruby to 2.3.3 and bundle update gems without changing outward facing interface.


Gemfile
```diff
@@ -21,25 +21,23 @@ gem 'net-ldap'
...
 gem 'puma'
-gem 'rack', '1.6.4' # Remove in rails 5
 gem 'rack-cors', :require => 'rack/cors'
 gem 'grape-middleware-lograge'
 gem "rack-timeout"
 
 gem 'grape-swagger'
 gem 'kaminari'
+gem 'kaminari-grape' #needed for kaminari 1.x
 gem 'grape-kaminari'
```
dmlond, line 31: do we need both kaminari-grape and grape-kaminari!??

Gemfile
```diff
@@ -21,25 +21,23 @@ gem 'net-ldap'
 
 # Use puma as the webserver in development
 gem 'puma'
-gem 'rack', '1.6.4' # Remove in rails 5
```
dmlond, line 24: just want to make sure that rack 2.0.1 doesnt have the same bug as 1.6.4, which was latest at the time of  [this commit](https://github.com/dmann/duke-data-service/commit/25d63bb8c431833611114f4f33e4d7915540af94) does not still have a bug. I wish I had better described the symptoms of the bug :(

dmann: Yeah, it's unfortunate that there isn't a test associated with that commit...

Gemfile
dmann: The newest release of grape-kaminari appears to need kaminari-grape in order to work correctly.

Gemfile
dmlond: :( it's hard to test rack. It will probably be fine :)

dmlond: I think it was something which prevented puma from starting in the docker container. If you can get the service running in docker, it is probably fine

dmann: I haven't noticed any docker issues when running rspec, but that doesn't kick off a puma server though.

dmlond (approved):
:1st_place_medal:

PR #979, opened 2017-04-15
DDS-870-create_or_update_indexes

dmlond:
elasticsearch_index_job now attempts to update, but creates if it doesnt exist

to prevent loss of data

job_shared_examples an ElasticsearchIndexJob tests this

### DIFF
--- a/spec/support/job_shared_examples.rb
+++ b/spec/support/job_shared_examples.rb
@@ -27,31 +27,40 @@ shared_examples 'an ElasticsearchIndexJob' do |container_sym|
   context 'update' do
     let(:existing_container) { FactoryGirl.create(container_sym) }
     let(:job_transaction) { described_class.initialize_job(existing_container) }
+    let(:original_name) { existing_container.name }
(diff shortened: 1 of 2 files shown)


spec/support/job_shared_examples.rb
```diff
@@ -29,13 +29,31 @@
...
-        expect(existing_container).to receive(:__elasticsearch__).and_return(mocked_proxy)
-        ElasticsearchIndexJob.perform_now(job_transaction, existing_container, update: true)
-      }
+
+      context 'index exists' do
+        it {
+          expect(existing_container).to be_persisted
+          mocked_proxy = double()
+          expect(mocked_proxy).to receive(:update_document)
+            .with(ignore: 404).and_return(true)
```
dmann: I think it would be better do to this as an integration test against Elasticsearch. This smells of "don't mock what you don't own".

dmann (changes requested):

dmann (approved):
👍

PR #982, opened 2017-04-20
DDS-725-fix_rake_graphdb

dmlond:
in [commit b9b4b00d8d91791b533357349c7b663f1a9a6ca0)[https://github.com/Duke-Translational-Bioinformatics/duke-data-service/commit/b9b4b00d8d91791b533357349c7b663f1a9a6ca0) this rake task was changed to only create

nodes for existing objects. It no longer automatically inferred associated_with and attributed_to
prov_relations from the audits. The rake task spec was not changed to match the new expectations.
Since it does not automatically run, it was not caught at the time. This fixes the spec to
meet the new expectations.


dmann (approved):
👍

PR #984, opened 2017-04-28
DDS-869 error queue handler

dmann:
The ErrorQueueHandler is a service which will primarily be used via a Rake task. It's purpose is to provide an interface to the Maxretry error queue, which allows non-destructive reading of its messages and the ability requeue jobs to the gateway exchange.


app/services/error_queue_handler.rb
```diff
@@ -0,0 +1,108 @@
...
+      republish_message(channel, msgs.last)
+      channel.ack(delivery_tags.pop)
+    end
+    msgs
+  end
+
+  def requeue_messages(routing_key:, limit: nil)
+    msgs = []
+    each_error_queue_message do |msg, channel, delivery_tags|
+      if routing_key && msg.first[:routing_key] == routing_key
```
dmlond: either your routing_key keyword argument needs a default (e.g. requeue_messages(routing_key: nil,...) or you can remove the extra test for routing_key because the method will always throw an ArgumentError if routing_key is not defined in the arguments.

dmlond (approved):
looks great! good work 💯

PR #987, opened 2017-05-02
added a section on Proposed Elasticsearch Index Strategy

dmlond:


mgardnerpsu (changes requested):
Amend with decision to index/return entire folder/file resource payload from Elastic and extend with tags/metadata.  This will eliminate need to make another trip to Postgres via ActiveRecord and provide clients with common JSON payload for file.

PR #993, opened 2017-05-08
Dds 872 eventual consistency status

mgardnerpsu:


dmlond (approved):
looks good. We should stress that error message codes will never change once in production.

dmann (commented):
👍

PR #996, opened 2017-05-10
DDS-876 sneakers configuration via env

dmann:
Include all available Sneakers and Bunny configuration in sneakers initializer.


dmlond (approved):
👍

PR #999, opened 2017-05-16
DDS-878 Recreate job_transactions table with UUID primary key

dmann:


dmlond (approved):
verified with my local original migration. looks good

PR #1004, opened 2017-05-18
DDS-880 bunny channel close bug

dmann:
Creating a class instance hash to store one Sneakers::Publisher per queue_name, which should amount to one RabbitMQ channel per ApplicationJob subclass. Hopefully this will stop the channel sprawl that we were seeing, where a new channel was opened (and not closed) for each enqueued job.


dmlond (approved):
🥇

PR #1005, opened 2017-05-18
Background Job bug fix (v1.3.13) to ua_test

dmann:
This addresses the RabbitMQ channel sprawl that resulted from new Sneakers::Publisher instances being created with each enqueued job. Publishers are now being cached on a per queue basis.


dmlond (approved):
👍

PR #1007, opened 2017-05-18
DDS-879 workflow sprawl script

dmann:
A workflow script for creating a project and populating it with a hierarchy folders, sprawling in both depth as well as breadth.


dmlond (approved):
👍

PR #1015, opened 2017-05-24
Background Jobs (v1.3.x) to Production

dmann:
Changes include:
- Using Sneakers gem to handle background jobs via RabbitMQ
- Background jobs for project storage initiation, elasticsearch indexing, and child deletions
- Gem upgrade, including Rails to version 5.x
- Ruby upgraded to 2.3.3


dmlond (approved):
👍

PR #1017, opened 2017-06-03
DDS-826-reworking_search

dmlond:
implements POST /api/v1/search/files_folders


mgardnerpsu (approved):

app/api/dds/v1/search_api.rb
```diff
@@ -76,12 +78,81 @@ class SearchAPI < Grape::API
...
+        end
+        optional :filters, type: Array, desc: 'List of boolean search objects (predicates) that are joined together by AND operator, represented as key: value pairs' do
+          optional :kind , type: Array, desc: 'Limit search context to list of kinds; valid options are dds-folder and dds-file - requires a set of comma-delimited kinds like so: {"kind": ["dds-file"]}. If not specified, defaults to all options.'
+          optional 'project.id', type: Array, desc: 'Limit search context to list of projects; requires a set of comma-delimited ids like so: {"project.id": ["345e...", "2e45..."]}. Filters out projects to which the user is not allowed access. If not specified, defaults to projects in which the current user has view permission.'
+        end
+        optional :aggs, type: Array, desc: 'List of aggs (aggregates) that should be computed.' do
+          optional :field, type: String, desc: 'The field for which you want to compute agg; valid options are project.name and tags.label. Required if one or more aggs are submitted.'
```
dmann, line 113: What does grape do when a `requires` is set inside of a `optional` parameter? If it works as expected, maybe we could get away from having "Required if one or more aggs are submitted." in the parameter descriptions.

app/api/dds/v1/search_api.rb
```diff
@@ -76,12 +78,81 @@ class SearchAPI < Grape::API
...
+      end
+      post '/search/folders_files', root: false do
+        authenticate!
+        search_params = declared(params, include_missing: false)
+        project_filter_exists = false
+
+        search_params[:filters] ||= []
+        search_params[:filters].each do |f|
+          if f.has_key? 'project.id'
+            filtered = policy_scope(Project).where(is_deleted: false, id: f['project.id']).all.to_a.map{|p| p.id}
```
dmann, line 139: The policy_scope calls are fairly complex and expensive. Since a non-specific error is raised as a result of this, maybe we could refactor this to collect the 'project.id' filters in an array and then flatten that and pass it to the policy scope.

app/api/dds/v1/search_api.rb
```diff
@@ -76,12 +78,81 @@ class SearchAPI < Grape::API
...
+        unless project_filter_exists
+          search_params[:filters] << {'project.id' => policy_scope(Project).where(is_deleted: false).all.map {|p| p.id}}
+        end
+
+        f = FolderFilesResponse.new
+          .filter(search_params[:filters])
+          .query(search_params[:query_string])
+          .aggregate(search_params[:aggs])
+          .post_filter(search_params[:post_filters])
+          .search
```
dmann, line 153: This message chaining looks really clean. A++  😃

app/models/folder_files_response.rb
```diff
@@ -0,0 +1,238 @@
...
+        query[:query][:filtered][:query][:query_string][:fields] = @query_string[:fields]
+      else
+        query[:query][:filtered][:query][:query_string][:fields] = @@supported_query_string_fields
+      end
+    end
+
+    if @aggs
+      query[:aggs] = {}
+      @aggs.each do |agg|
+        size = agg[:size] || 20
```
dmann: Would it be overkill to create a constant to hold the default agg size?

circle.yml
```diff
@@ -46,6 +46,7 @@ deployment:
     - heroku run rake --exit-code db:seed --app 'dukeds-dev'
     - heroku run --exit-code rake db:data:migrate --app 'dukeds-dev'
     - heroku run --exit-code rake elasticsearch:index:rebuild --app 'dukeds-dev'
+    - heroku run config:unset RECREATE_SEARCH_MAPPINGS
```
dmann, line 49: Nice! No more forgetting to unset that config var!

dmann (changes requested):
Looks good. I really like how the FolderFilesResponse model came out, very clean! Just a few minor changes/refactoring, but that's to be expected with a PR with over 2k lines of new code.

app/api/dds/v1/search_api.rb
dmlond: I tried this. It makes it difficult to provide the expected error payload for missing parameters, because it halts processing and returns a response before our code gets the chance to rescue from the exception.

app/api/dds/v1/search_api.rb
dmlond: There should only ever be at most one 'project.id', or 'kind' entry in the search_params[:filters] array. This will not call the policy_scope more than once per request. I could throw an exception for this in the FoldersFilesResponse.filter or validate_filters method to be sure.

app/models/folder_files_response.rb
dmlond: I can do that

app/api/dds/v1/search_api.rb
dmlond: I am asking Matt to make sure this is what he intended, and suggesting that filters and post_filters could be a Hash instead of Array.

dmlond: just got confirmation from Matt. He wants to accept an array of filters, so that we can extend our filters in the future to support objects with different schemas, and possible different predicates. BUT he is in agreement that we should only allow one 'project.id' or 'kind' object to be passed in. I am adding this to FolderFilesResponse and testing it.

app/api/dds/v1/search_api.rb
dmann: Makes sense, thanks for the explanation.

spec/models/folder_files_response_spec.rb
```diff
@@ -239,6 +261,24 @@
         }
       end
 
+      context 'duplicate kind' do
+        let(:duplicate_kind_filter) {
+          [
+            all_projects,
+            {"kind" => ["dds-folder"]},
+            {"kind" => ["dds-fille"]}
```
dmann: Is this a typo? Should this be "dds-file"?

dmlond: yes, I will fix it :)

dmann (approved):

PR #1024, opened 2017-06-12
DDS-872-eventual_consistency_status

dmlond:
attempt to push smaller commits for review
more to come


dmlond:
ready for official review :)

app/jobs/project_storage_provider_initialization_job.rb
```diff
@@ -4,6 +4,7 @@ class ProjectStorageProviderInitializationJob < ApplicationJob
   def perform(job_transaction:, storage_provider:, project:)
     self.class.start_job job_transaction
     storage_provider.put_container(project.id)
+    project.update(is_consistent: true)
```
dmann: Should handle a failed update here. Maybe call `update!` instead, which will throw an exception if a validation error occurs.

app/models/upload.rb
```diff
@@ -52,18 +55,28 @@ def manifest
...
+        UploadCompletionJob.perform_later(
+          UploadCompletionJob.initialize_job(self),
+          self.id
+        )
+        self
+      end
+    end
+  end
+
+  def make_consistent
```
dmann: Would it be possible to have a more descriptive name for this method? `make_consistent` is kind of vague IMHO.

app/models/upload.rb
```diff
@@ -52,18 +55,28 @@ def manifest
...
-            raise IntegrityException, "reported size does not match size computed by StorageProvider"
-          end
-          self
-        end
+      response = storage_provider.put_object_manifest(project_id, id, manifest, content_type, name)
+      meta = storage_provider.get_object_metadata(project_id, id)
+      unless meta["content-length"].to_i == size
+        raise IntegrityException, "reported size does not match size computed by StorageProvider"
       end
+      update_attributes({
```
dmann: `update_attributes` is an alias for `update`[1] and will run all the validations and whatnot on the object. I'm not sure what will happen here in the case of a validation failure.

1. http://api.rubyonrails.org/classes/ActiveRecord/Persistence.html#method-i-update_attributes

app/models/upload.rb
```diff
@@ -81,8 +94,8 @@ def integrity_exception(message)
     fingerprints.reload
     update_attributes({
```
dmann: Possibly another case for `update!`

app/serializers/upload_serializer.rb
```diff
@@ -1,16 +1,21 @@
...
 
+  def is_consistent
+    object.is_consistent?
+  end
+
   def status
     {
       initiated_on: object.created_at,
       completed_on: object.completed_at,
+      has_integrity_exception: object.has_integrity_exception?,
```
dmann: We have a couple issues here. First, the attribute `has_integrity_exception` isn't actually set anywhere, so this will always return `false`. That's fine, because I don't think that we actually need that attribute. The integrity of the upload can be inferred from the presence of the `error_message`, which is set when an integrity exception is raised.

If uploads start using `error_message` for non-integrity related exceptions, I suggest we add a `error_type` attribute.

dmann (changes requested):
Just a couple changes, nothing major though.

app/models/upload.rb
dmlond: ARGH! vcr! :{

dmann (approved):

PR #1026, opened 2017-06-19
DDS-862_app_status_errors

dmlond:
should work correctly now


Gemfile
```diff
@@ -72,14 +72,17 @@ group :development, :docker, :test do
   gem 'rspec-rails'
 end
 
+group :docker, :test do
```
dmann, line 75: Would it make sense to add `:development` to this group? I'm not sure how we'd use it from the rails console, since we wouldn't want to commit code with `binding.pry` in it... Hmm, maybe just ignore this thought then.  :P

app/api/dds/v1/app_api.rb
```diff
@@ -60,7 +59,7 @@ class AppAPI < Grape::API
               Sneakers::CONFIG[:retry_error_exchange],
               ApplicationJob.distributor_exchange_name
             ].each do |expected_exchange|
-              unless Sneakers::CONFIG[:connection].exchange_exists? expected_exchange
+              unless ApplicationJob.conn.exchange_exists? expected_exchange
```
dmann: :wrist_slap: ApplicationJob.conn is a private method...

Also, if the problem is that the connection isn't started, there should be a test that demonstrates that.  :P

app/api/dds/v1/app_api.rb
```diff
@@ -76,7 +75,7 @@ class AppAPI < Grape::API
               Sneakers::CONFIG[:retry_error_exchange]
             ].concat(application_job_workers)
             .each do |expected_queue|
-              unless Sneakers::CONFIG[:connection].queue_exists? expected_queue
+              unless ApplicationJob.conn.queue_exists? expected_queue
```
dmann: BTW, if you want ApplicationJob.conn to be public, you'll have to write some unit tests for it.

dmann (changes requested):
It would be good to have some tests to justify these changes...

dmlond:
wrote some tests

dmann (approved):

PR #1027, opened 2017-06-22
UploadCompletionJob workers and rake tasks

dmlond:
registered in jobs_runner
specs test presence of registered jobs_runner, and rake task

### DIFF
--- a/spec/lib/tasks/workers_rake_spec.rb
+++ b/spec/lib/tasks/workers_rake_spec.rb
@@ -39,6 +39,15 @@ describe "workers" do
     it_behaves_like 'a queued job worker', :expected_job_class
   end
 
+  describe 'workers:complete_upload:run' do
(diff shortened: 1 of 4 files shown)


spec/lib/tasks/workers_rake_spec.rb
```diff
@@ -39,6 +39,15 @@
     it_behaves_like 'a queued job worker', :expected_job_class
   end
 
+  describe 'workers:complete_upload:run' do
```
dmann: This, as well as the above four describe blocks, are not needed since the `JobsRunner.workers_registry.each` loop below automatically tests all the individual worker tasks. Sorry for the confusion this legacy code has created!

dmann (changes requested):
Looks good. Just remove the redundant tests.

dmann (approved):

PR #1028, opened 2017-06-23
DDS-355 remove DataFile upload association

dmann:


app/models/user.rb
```diff
@@ -22,9 +22,9 @@ class User < ActiveRecord::Base
     -> { where(is_deleted: false) },
     class_name: 'DataFile',
     through: :permitted_projects,
-    source: :data_files,
-    foreign_key: "creator_id"
-  has_many :uploads, through: :created_files
+    source: :data_files
+  has_many :file_versions, through: :created_files
+  has_many :uploads, through: :file_versions
```
dmlond: should use the foriegn_key: "creator_id" to limit to just that user's uploads

app/models/data_file.rb
```diff
@@ -26,6 +24,14 @@ class DataFile < Container
 
   delegate :http_verb, to: :upload
 
+  def upload=(val)
+    @upload=val
```
dmlond: I have a great idea! Why dont we get rid of the callbacks for setting the current_version, and just use the upload setter to set the current file version :P

dmlond (changes requested):

dmlond (approved):
ampersands for the win !!!

PR #1030, opened 2017-06-29
DDS-901_consistency_db_data_migrate

dmlond:


spec/lib/tasks/db_data_migrate_rake_spec.rb
```diff
@@ -141,5 +141,66 @@
...
+          invoke_task
+          expect(record_class.where(is_consistent: nil)).not_to exist
+          record.reload
+          expect(record.is_consistent).to eq true
+        end
+      end
+
+      context 'record is not consistent' do
+        it 'should update is_consistent to false' do
+          invoke_task
```
dmann: Is there a reason that you're not checking `record_class.where(is_consistent: nil)` like you're doing in the above it block?

spec/lib/tasks/db_data_migrate_rake_spec.rb
```diff
@@ -141,5 +141,66 @@
...
+
+    shared_examples 'a consistency migration' do |prep_method_sym|
+      let(:record_class) { record.class }
+      before do
+        expect(record).to be_persisted
+      end
+      context 'record is consistent' do
+        let(:prep_method) { send(prep_method_sym) }
+        before do
+          prep_method.call
```
dmann: Using a `Proc` and having to use `.call` doesn't really fit any of our existing patterns. I believe you could get the same result without a `Proc` and by replacing the above with: `expect { proc_method }.not_to raise_error`

Which gives the added bonus of trapping possible exceptions when `prep_method` is executed.

spec/lib/tasks/db_data_migrate_rake_spec.rb
```diff
@@ -141,5 +141,66 @@
...
+      before do
+        expect(storage_provider).to be_persisted
+      end
+
+      context 'for project' do
+        let(:record) { FactoryGirl.create(:project, is_consistent: nil) }
+        let(:init_project_storage) {
+          Proc.new do
+            storage_provider.put_container(record.id)
+            expect(storage_provider.get_container_meta(record.id)).to be
```
dmann: I've been trying to get away from using `.to be` without an argument. It's kinda vague and has lead to confusion when going over the tests with new developers. If you're trying to assert that something is returned, use `.not_to be_nil` instead.

spec/lib/tasks/db_data_migrate_rake_spec.rb
```diff
@@ -141,5 +141,66 @@
...
+        }
+        it_behaves_like 'a consistency migration', :init_project_storage
+      end
+
+      context 'for upload' do
+        let(:record) { FactoryGirl.create(:upload, is_consistent: nil, storage_provider: storage_provider) }
+        let(:init_upload) {
+          Proc.new do
+            storage_provider.put_container(record.project.id)
+            expect(storage_provider.get_container_meta(record.project.id)).to be
```
dmann: See above regarding `.to be`

spec/lib/tasks/db_data_migrate_rake_spec.rb
```diff
@@ -141,5 +141,66 @@
...
+            record.create_and_validate_storage_manifest
+            record.update(is_consistent: nil)
```
dmann: Be aware that using `update` will run validations. If you want to skip validations, use `update_attribute`. If you also want to skip callbacks, use `update_column`. If you don't care about either, use `update!` so that an exception is thrown if something doesn't work as expected.

spec/lib/tasks/db_data_migrate_rake_spec.rb
```diff
@@ -141,5 +141,66 @@
...
+            expect(storage_provider.get_object_metadata(record.project.id, record.id)).to be
```
dmann: See above regarding `.to be`

dmann (changes requested):
Looks good for the most part, just a few changes.

spec/lib/tasks/db_data_migrate_rake_spec.rb
dmlond: no :)

dmlond:
ready for final review

spec/lib/tasks/db_data_migrate_rake_spec.rb
dmann: I meant on both sides of the `invoke_task` call...  :P

dmlond: ok. should be good to go

dmann (approved):

PR #1033, opened 2017-07-13
Rails ERD Changes. Builds model diagram

bchaballout:
This is how you run to build the model diagrams: `docker-compose run bundle exec erd`

More information about Rails ERD available here:
http://voormedia.github.io/rails-erd/install.html

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -86,7 +86,7 @@ group :test do
   gem 'bunny-mock'
 end
 
-group :development do
+group :docker do
   gem "rails-erd"
 end
 
--- a/Gemfile
+++ b/Gemfile
@@ -86,6 +86,10 @@ group :test do
   gem 'bunny-mock'
 end
 
+group :development do
+  gem "rails-erd"
+end
+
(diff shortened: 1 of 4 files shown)


Gemfile
```diff
@@ -86,6 +86,10 @@ group :test do
   gem 'bunny-mock'
 end
 
+group :development do
```
dmann: change this to group :docker
--dmlond

docker-compose.dev.yml
```diff
@@ -37,6 +37,17 @@ services:
     volumes_from:
       - localdev
     entrypoint: ['bundle']
+    links:
```
dmann: We should create a new application entry using `bundle` as an example, changing the name and entry point to `erd`. This should allow us to call `docker-compose run erd` without having to use `bundle exec`.

dmann (changes requested):

dmlond (approved):
looks great!

dmann (approved):

PR #1034, opened 2017-07-17
DDS-915 MessageLog queue testing bug

dmann:
Elasticsearch query page size set to 1000 when testing.
Also created an expectation that checks that the search hits returned equals the hit total.


Donmurry (approved):
👍

PR #1041, opened 2017-07-31
DDS-922_optimize_child_deletion_job

dmlond:
not ready for review yet


dmlond:
the commit message for [c7350a3](https://github.com/Duke-Translational-Bioinformatics/duke-data-service/pull/1041/commits/c7350a32556b4e6e4d5b5407848f7326ddd2d709) should read child_minder now implements #delete_children instead of #child_deletion.  The method is actually #delete_children

spec/support/concerns/child_minder_shared_examples.rb
```diff
@@ -132,24 +140,62 @@
     end
   end
 
-  describe '#delete_children' do
+  describe '#paginated_children' do
```
dmann: Since you're only using it within ChildMinder, I think you could probably make #paginated_children a private method and then you wouldn't have to test it directly.

dmann (changes requested):

dmann (approved):

PR #1050, opened 2017-08-11
Initial API design for upload state model

mgardnerpsu:


dmann (changes requested):
Let's go with Design Option A. I'm not sure how we normally handle this, but I think it would make sense to update the design doc to remove the extraneous Design Option B.

caseychoiniere (commented):
Agree with DM. Option A.

PR #1054, opened 2017-08-17
DDS-924 rule change update file permission

dmann:
- File updates no longer restricted to the upload creator
- File rename was using the incorrect permission. Changed from :create_file to :update_file, which is what is currently specified in the api docs.


dmlond (approved):
👍

PR #1056, opened 2017-08-17
DDS-925 rule change delete own project permission

dmann:
…oject_admin exists for the project


dmlond (approved):
siblings concept is nice. Looks good.

PR #1065, opened 2017-09-05
Removed global trash search - start with project trash for initial im…

mgardnerpsu:
…pl; Factored out user/project stats endpoints - will migrate to stats specification.


dmlond (approved):
👍

PR #1069, opened 2017-09-08
Initial draft to fix too many chunks uploaded issue.

mgardnerpsu:


api_design/DDS-974-enforce_upload_chunk_size.md
```diff
@@ -0,0 +1,54 @@
+# DDS-974 Enforce Upload Chunk Size
+
+## Deployment View
+
+N/A
+
+## Logical View
+
+A critical issue was recently exposed with the chunked upload capability.  The underlying storage system (Swift), can only coalesce a large file that is <= 1000 segments (i.e. referred to as chunks in the DDS API).  To remedy this situation, the API will be extended to enforce a minumum chunk size for chunked file uploads.
```
dmlond: we should probably avoid mentioning 1000 specifically, as it might confuse others into thinking that this cannot be changed. Maybe say something to the effect that each storage provider is configured with a set maximum number of segments, beyond which it cannot coalesce a large file.

dmlond (changes requested):

api_design/DDS-974-enforce_upload_chunk_size.md
mgardnerpsu: Good point - I will update now and push

api_design/DDS-974-enforce_upload_chunk_size.md
```diff
@@ -0,0 +1,54 @@
...
+|Endpoint |Description |
+|---|---|
+| `POST /projects/{id}/uploads` | Intitate a chunked upload. |
+| `PUT /uploads/{id}/chunks` | Generate and return a pre-signed upload URL for a chunk.  |
+
+#### API Specification
+This section defines the proposed API interface extensions.
+
+##### Intitate a chunked upload
+`POST /projects/{id}/uploads`
```
dmlond: should we return an error if the user attempts to initiate an upload of a file that is too large (max_segments * max_chunk_upload_size)?

mgardnerpsu: yes - and do we want to extend storage provider endpoint to return the max segments, min size, max size, that we support at API level?  i assume you would keep these settings at this level - may be good to return them for informational purposes as well.

dmlond: If you can document these proposed changes to the StorageProvider in the api_design doc, that should set it up for us to start implementation.

api_design/DDS-974-enforce_upload_chunk_size.md
```diff
@@ -0,0 +1,54 @@
...
+	"reason": "Invalid chunk size specified, must be in range {min}-{max}"
+	"suggestion": "Use valid chunk size"
+}
+```
+	
+## Implementation View
+
++ The offcial GCB python client and DDS Web portal client will need to be modifed to interface with these chunked upload API extensions.
+
++ A migration will be needed to remedy any existing uploads that have been marked as completed by the client (i.e. `status.completed_on != null`) and have greater that 1000 chunks (i.e. Swift segments).  This may be accomplished by increasing the `max_manifest_segements` on the Swift instance.
```
dmlond: I dont think this is true anymore. Once we up the current slo max_segments to 2000 in production, all current uploads should complete without issue. I do not think that we will be dropping that back down. If we do, then we can come up with a migration.

dmlond (approved):
👍

api_design/DDS-974-enforce_upload_chunk_size.md
```diff
@@ -0,0 +1,89 @@
...
+###### Response Headers (Extensions)
+The following custom response headers will be added to inform clients of the minimum chunk size that may be utlized to ensure chunks can be coalesced, as well as the maximum chunk size the storage provider can accommodate.
+
++ **x-min-chunk-upload-size** - The minimum chunk size in bytes.
++ **x-max-chunk-upload-size** - The maximum chunk size in bytes.
+
+##### Generate and return a pre-signed upload URL for a chunk
+`PUT /uploads/{id}/chunks`
+
+###### Response Messages (Extensions)
```
dmann, line 58: We'll need one more 400 response for when `PUT /uploads/{id}/chunks` tries to create a chunk on an upload that has `chunk_max_number` of chunks.

dmann (commented):
Please add a response for when `chunk_max_number` is exceeded.

PR #1072, opened 2017-09-12
Enforce Max Chunk Uploads

dmlond:
migrated to include max_chunk_number and max_chunk_size_bytes
validates presence of these
factory creates these
swift.local.env sets these for local swift instantiation


dmann (approved):
👍

PR #1080, opened 2017-09-27
tag elasticsearch index management

dmlond:
tags should index/update their taggable elasticsearch indices the same way that meta_properties index/update their meta_template.templatable objects


dmann (approved):
👍

PR #1082, opened 2017-09-28
DDS-1009 custom serialization via params

dmann:


app/api/dds/v1/children_api.rb
```diff
@@ -14,6 +14,9 @@ class ChildrenAPI < Grape::API
       end
       params do
         optional :name_contains, type: String, desc: 'list children whose name contains this string'
+        optional :exclude_response_fields, type: Array[String],
```
dmlond, line 17: you should run launch_application and view the /apiexplorer to see if/how this ends up in swagger.

spec/requests/children_api_spec.rb
```diff
@@ -44,6 +44,20 @@
           deleted_folder,
           other_folder
         ] }
+
```
dmlond, line 47: can this be turned into a shared_example?

dmlond (changes requested):
looks good, just a few suggestions

app/api/dds/v1/children_api.rb
dmann: Na, I don't feel like messing with `launch_application`. If after deploying to ua_test we find that it's not usable in swagger, I'll look into other solutions.

spec/requests/children_api_spec.rb
dmann: Yeah, I was waffling between shared_context and shared_examples for this bit. I landed on not refactoring due to decision paralysis. If this feature sticks around and is expanded to other endpoints, I'll definitely refactor it and put it with the other request shared examples.

dmann:
I will merge this into my main task branch and create a new PR to ua_test from there.

PR #1084, opened 2017-10-03
DDS-997-998-is_deleted_index

dmlond:
- adds is_deleted to the mapped folder and file elasticsearch indices
- get /folder_files now only returns objects where is_deleted: false
- Added implementation notes to, and updated status in the design document


mgardnerpsu (approved):

dmann (approved):
👍

Issue #1071, opened 2017-09-08
Old project names show in search results and aggregations


caseychoiniere:
Creating an issue based on the discussion here https://dukefrds.slack.com/archives/C0AN58F0A/p1504651068000093

Searching for a file using `POST/search/folders_files` the results seem to include old project names (the project has since been renamed but the old name shows in the indexed results). This messes with the aggregate buckets. The results below show 17 docs matching in a project called `Directory Uploads!!!!!123` and 1 doc in a project called `Directory Uploads!!!!`. The name of the project was renamed and is no longer  `Directory Uploads!!!!!123`. All 18 of the documents actually reside in a project called `Directory Uploads!!!!`.  This also affects other result metadata such as the `ancestors` and `project` properties for files or folders.

```"buckets": [
        {
          "key": "Directory Uploads!!!!!123",
          "doc_count": 17
        },
        {
          "key": "Directory Uploads!!!!",
          "doc_count": 1
        }
      ]```

jonturneratduke:
Yep, this is a problem. We’ll have to re-build the search index on a periodic basis (knowing results will get stale over time) or trigger an add/delete to the search index when anything that affects the name happens. It’s the difference between keeping a ledger and doing an inventory. For accuracy’s sake, I prefer the track-the-changes-as-we-go approach. Curious what others think, esp. Matt?

caseychoiniere:
I think #1067 is equally important. As a user I would assume that if I added tags I would then be able to find my files via those tags.

caseychoiniere:
Also +1 for the track-the-changes-as-we-go approach

PR #1086, opened 2017-10-05
DDS-999-project_name_elasticsearch_index

dmlond:
This will eventually update folder.children ancestors when folder name changes as well.


dmann (approved):
This will get us off the ground. We need to upgrade to Elasticsearch 5.x and change this to use `update_by_query` to improve performance.

caseychoiniere:
@dmann or @dmlond  is this supposed to fix #1071 ? If so, it looks like old project names are still indexed.

PR #1088, opened 2017-10-10
Merge v1.3.28 & v1.3.29 into production

dmann:
- Allow json response customization for `children` endpoints via the `exclude_response_fields` parameter
- Remove minimum chunk size validation


Donmurry (approved):
👍

PR #1093, opened 2017-10-20
DDS-1025_elasticsearch_rake

dmlond:
creating MetaProperty mappings caused the index creation function to end in error, which
caused the rake task to end prematurely

test needed to search for file/folder by name instead of id, since we no longer index id


dmann (approved):
🔢

PR #1096, opened 2017-10-30
DDS-1024 graph db background job

dmann:
Refactoring the GraphDB code and creating a background job.


spec/support/concerns/graphed_shared_examples.rb
```diff
@@ -95,52 +114,36 @@
...
-      expect(graphed_relation.model_kind).to eq(subject.kind)
-      expect(graphed_relation.from_node.model_id).to eq(from_model.id)
-      expect(graphed_relation.to_node.model_id).to eq(to_model.id)
-    end
+  describe '#graph_model_class' do
+    it { is_expected.to respond_to :graph_model_class }
+    it { expect(subject.graph_model_class).to eq graph_model_class }
+  end
+
+  it 'should auto_create' do
```
dmlond: since none of these actually 'auto_create' it might be better to change this description to something different, or just make it a bare it { ... } block.

dmlond (dismissed):

dmlond (commented):
💯

spec/support/concerns/graphed_shared_examples.rb
dmann: Actually, they *all* auto create, which is why I changed all the code to remove the unused `auto_create` parameter.

dmlond: ok

dmlond (approved):
🥇

PR #1099, opened 2017-11-03
DDS-829_purgable_containers

dmlond:
DataFileSpec expects DataFile to behave like a Restorable and Purgable ChildMinder

 - requires alias children to file_versions

Refactored Shared Examples for Restorable ChildMinder, Purgable ChildMinder, and
UnRestorable ChildMinder to meet new expectations.

DataFileSpec and DataFilesApiSpec no longer expects DataFile to delete all of its
file_versions after it is deleted. Instead, this is taken care of by the ChildMinder
using background jobs.


spec/support/concerns/restorable_shared_examples.rb
```diff
@@ -135,84 +141,87 @@
...
+    if child_minder_children_sym
+      let(:child_job_transaction) { ChildRestorationJob.initialize_job(child_minder_children.first) }
+    end
+
+    let(:page) { 1 }
 
-      after do
-        Rails.application.config.max_children_per_job = @old_max
+    before do
+      expected_children.each do |expected_child|
```
dmann: It would be good to check that `expected_children` is not empty before calling `#each` on it. If it's empty, a lot of the expect statements in this section will be silently skipped. There are probably other places that could benefit from this sort of check.

dmann (changes requested):
Just the one thing, the rest looks good.

dmann (approved):

PR #1101, opened 2017-11-08
Dds 1029 trashbin endpoints

dmlond:


app/api/dds/v1/trashbin_api.rb
```diff
@@ -0,0 +1,129 @@
...
+          optional :kind, type: String, desc: 'Parent Kind'
+          optional :id, type: String, desc: 'Parent UUID'
+        end
+      end
+      put '/trashbin/:object_kind/:object_id/restore', root: false do
+        authenticate!
+        parent_params = declared(params, {include_missing: false}, [:parent])
+        object_kind = KindnessFactory.by_kind(params[:object_kind])
+        purge_object = object_kind.find_by!(id: params[:object_id], is_deleted: true)
+        unless purge_object.class.include? Restorable
```
dmann: I feel like this check should happen using `object_kind` on the line above the call to `.find_by!`. Also, the error_json formatting seems like it belongs in a separate helper method, which would help to reduce the clutter in this endpoint.

app/api/dds/v1/trashbin_api.rb
```diff
@@ -0,0 +1,129 @@
...
+          error_json = {
+            "error" => "404",
+            "code" => "not_provided",
+            "reason" => "#{purge_object.class.to_s} Not Found",
+            "suggestion" => "#{purge_object.kind} is not Restorable"
+          }
+          error!(error_json, 404)
+        end
+        unless purge_object.is_purged?
```
dmann: Nothing happens if `purge_object.is_purged?` is true? Seems like an immutable validation error or something should result from this...

app/api/dds/v1/trashbin_api.rb
```diff
@@ -0,0 +1,129 @@
...
+          purge_object.is_deleted = false
+          if purge_object.is_a?(FileVersion)
```
dmann: Could this logic be stored in the model? Possibly a `#restore` method?

app/api/dds/v1/trashbin_api.rb
```diff
@@ -0,0 +1,129 @@
...
+      end
+      params do
+        requires :object_kind, type: String, desc: 'Object Kind'
+        requires :object_id, type: String, desc: 'Object UUID'
+      end
+      put '/trashbin/:object_kind/:object_id/purge', root: false do
+        authenticate!
+        object_kind = KindnessFactory.by_kind(params[:object_kind])
+        purge_object = object_kind.find_by!(id: params[:object_id], is_deleted: true)
+        unless purge_object.class.include? Restorable
```
dmann: My comments for the `/restore` endpoint apply to the `/purge` endpoint as well.

dmann (changes requested):
The grape endpoints need to be DRYed out some, probably by moving logic to the model layer and possibly raising/rescuing exceptions instead of nested if statements. All the rest looks good.

dmlond:
that was useful. Thinking about these code review suggestions forced me to realize that I had missed the fact that containers could be restored to project roots, among some other things. Please review again :)

app/models/data_file.rb
```diff
@@ -103,4 +103,11 @@ def creator
     return unless create_audit
     create_audit.user
   end
+
+  def restore(child)
+    raise TrashbinParentException.new("#{kind} #{id} is deleted, and cannot restore its versions.::Restore #{kind} #{id}.") if is_deleted?
```
dmann, line 108: Some of this error message formatting could go in the exception classes. Maybe pass them `kind` and `id`, or an object?

app/api/dds/v1/trashbin_api.rb
```diff
@@ -0,0 +1,134 @@
...
+        }
+        error!(error_json, 404)
+      end
+      put '/trashbin/:object_kind/:object_id/restore', root: false do
+        authenticate!
+        parent_params = declared(params, {include_missing: false}, [:parent])
+        object_kind = KindnessFactory.by_kind(params[:object_kind])
+        purge_object = object_kind.find_by!(id: params[:object_id], is_deleted: true)
+        raise UnRestorableException.new(purge_object.kind) unless purge_object.class.include? Restorable
+        unless purge_object.is_purged?
```
dmann, line 83: I still don't know what to think about checking `.is_purged?` when restoring and not throwing an error or something. If it's gotten this far, `purge_object` is `is_deleted: true`, so if it's also `.is_purged?`, then trying to restore it shouldn't work, right?

dmann (commented):
Lots of good improvements. I think the exception classes could hold some formatting logic, maybe taking the calling object as a reference. Also see my comment regarding `/restore`. 👍

app/api/dds/v1/trashbin_api.rb
dmlond: I prefer keeping the check for is_purged?. It is a guard clause that quickly skips further SQL calls to get a parent or perform authorize!.

app/models/data_file.rb
dmlond: This is nice, but I think it can wait until the hardening phase in a couple of weeks. It isnt something that prevents code from moving forward.

dmlond:
Merging with the intent to refactor exceptions in a future hardening phase
[DDS-1035](https://jira.oit.duke.edu/browse/DDS-1035)

PR #1105, opened 2017-11-17
DDS-1039-grape_logging

dmlond:
uses [grape_logging](https://github.com/aserafin/grape_logging)

Also fixes the rake storage_provider:create task to set chunk_max_number of chunk_max_size_bytes


dmlond:
wrong target base :(

PR #1106, opened 2017-11-17
DDS-1039-grape_logging

dmlond:
uses (grape_logging)[https://github.com/aserafin/grape_logging] which restores log information about response status, timing, path, and params

also fixes rake storage_provider:create to set chunk_max_number of chunk_max_size_bytes


Donmurry (approved):
👍

PR #1107, opened 2017-11-21
DDS-1039-analytics

dmlond:
workflow.stress.sh uses workflow.sprawl.sh and new upload_file.sh to create a project and defined number of child folders and files in the DDSURL, using the DDSTOKEN, and print usage and children output.

This can be used to analyze timing of various endpoint calls.

### DIFF
--- a/workflow/workflow.sprawl.sh
+++ b/workflow/workflow.sprawl.sh
@@ -174,4 +174,4 @@ project_id=$(get_id_from_response) #`uuidgen` #`echo ${response} | jq -r '.id'`
 
 folder_total=0
 build_folders $project_id "dds-project" "${folders[@]}"
-echo "Created `echo $folder_total | bc` folders"
+echo "Created `echo $folder_total | bc` folders in project ${project_id}"
--- a/workflow/workflow.sprawl.sh
+++ b/workflow/workflow.sprawl.sh
@@ -1,10 +1,7 @@
 #!/bin/bash
-which jq > /dev/null
-if [ $? -gt 0 ]
-then
-  echo "install jq https://stedolan.github.io/jq/"
-  exit 1
-fi
+workflow_dir=`dirname $0`
+declare -r DIR=$(cd "${workflow_dir}" && pwd)
+source $DIR/restlib.sh
 
 max_folder_name_count=6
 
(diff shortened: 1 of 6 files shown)


workflow/workflow.sprawl.sh
```diff
@@ -1,10 +1,7 @@
 #!/bin/bash
-which jq > /dev/null
-if [ $? -gt 0 ]
-then
-  echo "install jq https://stedolan.github.io/jq/"
-  exit 1
-fi
+workflow_dir=`dirname $0`
+declare -r DIR=$(cd "${workflow_dir}" && pwd)
```
dmann, line 3: Slick.

workflow/workflow.stress.sh
```diff
@@ -0,0 +1,193 @@
...
+
+sprawl_command="${workflow_dir}/workflow.sprawl.sh"
+if ${pretend}
+then
+  sprawl_command="${sprawl_command} -p"
+fi
+
+if ${force}
+then
+  sprawl_command="${sprawl_command} -f"
```
dmann, line 151: FYI, it is possible to combine `-p` and `-f`. My use case was to see just how many files would be created when forcing sprawl to go over the `$max_folder_name_count`.

workflow/workflow.sprawl.sh
```diff
@@ -174,4 +126,4 @@ project_id=$(get_id_from_response) #`uuidgen` #`echo ${response} | jq -r '.id'`
 
 folder_total=0
 build_folders $project_id "dds-project" "${folders[@]}"
-echo "Created `echo $folder_total | bc` folders"
+echo "Created `echo $folder_total | bc` folders in project ${project_id}"
```
dmann, line 129: This is a useful addition. 👍

workflow/workflow.stress.sh
```diff
@@ -0,0 +1,193 @@
...
+workflow_dir=`dirname $0`
+declare -r DIR=$(cd "${workflow_dir}" && pwd)
+source $DIR/restlib.sh
+
+max_file_count=6
+
+usage_and_exit()
+{
+  read -d '' usage << USAGE
+usage: workflow.stress.sh -n files_per_folder [-hvp] [-d seconds]
```
dmann: The `-N` and `-f` options in this usage line.

dmann (changes requested):
Just fix the usage docs in the `workflow.stress.sh` script.

dmann (approved):

PR #1108, opened 2017-11-24
Chunk no longer updates Upload.etag on save

dmann:
This should fix the race conditions and db deadlock errors caused by multithreaded chunk creation.


dmlond (approved):
This is a good step. There are other places we need to review and remove this kind of parent update that can lead to race conditions, etc. 👍

PR #1112, opened 2017-11-30
neo4j changes

dmlond:
- updated neo4j in docker-compose and circle.yml to 3.1.6 to match graphenedb provisioned instance
- added config.neo4j.wait_for_connection = true to application.rb to ensure that the application
  waits for up to 60 seconds before crashing while neo4j starts
- replaced GRAPHSTORY_URL with GRAPHENEDB_URL in places where it is set or used


dmann (approved):
👍

PR #1119, opened 2017-12-15
DDS-1023 elasticsearch 6x upgrade

dmann:


dmann:
It doesn't look like upgrading to ES6 is going to work. We can try again in the future, when the support is hopefully better.

PR #1121, opened 2017-12-20
DDS-1053_elasticsearch_reindex

dmlond:
implements versioned elasticsearch indices aliased to standard elasticsearch indices
    - includes mapping and migration version
    - these are hard-coded into the models
  implements ElasticsearchHandler service that contains reusable elasticsearch methods that were
  formerly only in the rake task.

ElasticsearchHandler implements smart_reindex
    - if the migration version has not changed, will fast reindex data from the old version to the new version
    - if the migration_version has changed, requires a manual, slow reindex


app/models/activity.rb
```diff
@@ -2,7 +2,25 @@ class Activity < ActiveRecord::Base
...
+
   include SearchableModel
+  # change this to a new uuid (lowercase letters!) any time
+  #  - a migration is created to add/remove fields
+  #    and its serializers (standard and search)
+  #  - relationships are added to/removed from the serializers
+  @@migration_version = '429d08e4-622d-456a-a87d-84c288857320'
+
+  # change this variable to a new uuid (lowercase letters!) any time the mappings below change
+  @@mapping_version = 'bab2183a-5ed5-4cd3-b05e-916e20817dd7'
```
dmann: Not sure we need to use class variables for these versions. I think we could just place the strings directly in the methods.

spec/support/concerns/searchable_model_shared_examples.rb
```diff
@@ -5,6 +5,19 @@
   it { expect(described_class).to include(JobTransactionable) }
   it { expect(described_class).to include(Elasticsearch::Model) }
   it { expect(described_class).not_to include(Elasticsearch::Model::Callbacks) }
+  it { expect(FolderFilesResponse.indexed_models).to include described_class }
+
+  describe 'elasticsearch migrations' do
+    it { expect(described_class).to respond_to(:index_name) }
+    it { expect(described_class).to respond_to(:mapping_version) }
+    it { expect(described_class).to respond_to(:migration_version) }
```
dmann, line 13: Should test that the *_version methods don't raise NotImplementedError.

app/services/elasticsearch_handler.rb
```diff
@@ -0,0 +1,181 @@
...
+        end
+      end
+    end
+  end
+
+  def drop_indices(client=Elasticsearch::Model.client)
+    client.indices.delete index: '_all'
+  end
+
+  def smart_reindex_indices
```
dmann: This method is 42 lines long, with `if` statements nested 4 levels deep. It /should/ be refactored, but in the interest of time, we can pair tomorrow and walk through the code together.

dmann (commented):
I'm not finished reviewing. There's a lot here and I think it would be more efficient to go through it together.

spec/support/concerns/searchable_model_shared_examples.rb
dmlond: the test that includes line 16 will fail if a NotImplementedError is raised

dmann (approved):
👍

PR #1128, opened 2018-01-11
refactored elasticsearch_handler

dmlond:
ElasticsearchHandlerSpec expects ElasticsearchHandler
  - start_scroll
    - sets a default size in the body to 10000
    - sets the current_scroll_id from the search response
      before returning the response
  - next_scroll
     - no longer takes a scroll_id parameter
     - submits a scroll search with the current_scroll_id parameter
     - sets the current_scroll_id parameter from the search response
       before returning the response
  - fast_reindex
     - no longer returns a boolean to indicate success
     - implements an attr_reader :has_errors which is true if fast_reindex
       encounters errors
     - implements restart logic
       - stores the last_scroll_id in an attr_reader
       - takes a boolean restart parameter, if false, starts a new scroll
         if true, gets the next_scroll with the current_scroll_id
  - smart_reindex
     - uses new fast_reindex and has_errors functionality
     - will retry 4 attempts when
       Elasticsearch::Transport::Transport::Errors::GatewayTimeout
       exceptions occur

ElasticsearchHandler implements spec

ElasticsearchRakeSpec expects ElasticsearchRake to use new
ElasticsearchHandler functionality

### DIFF
--- a/app/services/elasticsearch_handler.rb
+++ b/app/services/elasticsearch_handler.rb
@@ -9,9 +9,10 @@
 #    verbose is true, some methods will print information to stderr
 #
 class ElasticsearchHandler
-  attr_reader :verbose
+  attr_reader :verbose, :has_errors, :current_scroll_id
   def initialize(verbose:false)
     @verbose = verbose
+    @has_errors = false
   end
 
   # create_indices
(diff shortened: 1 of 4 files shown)


dmann (approved):
👍

PR #1133, opened 2018-02-08
DDS-1061 optimized directory structure endpoint

dmlond:


johnbradley:
The fields DukeDSClient is currently using from a file response are:
```JSON
{
  "kind": "dds-file",
  "id": "4b24c20b-a4ed-4910-b51a-b747c76c4518",
  "parent": {
    "kind": "dds-project",
    "id": "dd3673ff-b011-4fe4-b843-5470c26938e1"
  },
  "name": "here.pyc",
  "current_version": {
    "id": "3217897d-8668-4012-9c11-59d1ad38356e",
    "upload": {
      "size": 5843,
      "hashes": [
        {
          "algorithm": "md5",
          "value": "866aecf23d71d084133007b6122f1be5"
        }
      ]
    }
  }
}
```

In other words we do not need `audit` or `current_version.hashes.audit`, 'ancestors' or 'project'.
Currently we set `exclude_response_fields` to  'audit', 'ancestors', 'project'.
For remote folders we need the same details only there is no `current_version` part.

The information we need is:
- the full remote path to the object (parents' names and file's name)
- the current version size and MD5 hash
- an id that we can use for fetching the current file url (file id or current version file id)

This allows us to:
- know where to download the file to
- if there is an existing file, is it different from the remote file
- get a url for downloading the file contents

dmlond:
thanks. This is good information.
This endpoint will most likely not support `name_contains`. Do you see any issues with not having this ability to filter on names?

dmlond:
Please let us know if any of the Proposal bullets raise concerns.

johnbradley:
> Do you see any issues with not having this ability to filter on names?

No. We do not currently use `name_contains` for uploading and downloading projects.

## Responding to bullet points from proposal

> - determine minimal query parameters required for their needs, for example, this endpoint may not need to support `name_contains` or `exclude_response_fields`

I do not think we will need either feature.


> - include deleted children in response documents to facilitate detection of changes in the overall filesystem

Currently we only perform upload and download. There is no sync functionality in DukeDSClient. This might be a nice enhancement down the road if we implement a `sync` command.


> - sort children in order of last_updated_on to ensure most recently changed objects come out first in responses

I don't know that the order of the files and folders in the response matters that much when downloading. We print out filenames with the % of progress, but I don't know that our users have a particular order in mind for downloading the files. They are just going to wait until they are all downloaded. We currently do some re-arranging of the order so we create the directories before we download the files into the directories.

api_design/DDS-1061-optimized_directory_structure_endpoint.md
```diff
@@ -0,0 +1,155 @@
...
+
+###### Deployment Requirements
+
+## Logical View
+
+The goal of this development epic is to produce a set of endpoints to be used in place of the `/projects/:id/children` and `/folders/:id/children` endpoints.
+
+#### Background
+
+The legacy children endpoints have proven to be suboptimal for the GCB clients needs. The GCB client needs a paginated endpoint to list ALL children of a project or folder. The current children endpoints meet the information needs for their use, but it currently takes on average 15 minutes to get the directory structure for a project or folder with many subfolders and files (similar to our ua_test sprawl project).
```
dleehr: The use case for the GCB client we have is uploading and downloading multiple files, usually entire projects. While the deliverable for such a use case might end up being a "paginated endpoint to list ALL children of a project or folder", I wouldn't say that's we're specifically asking for that.

dmlond: That is definitely a goal of this part of the feature development process, to get an exact spec of your needs. 

We will change the wording of the Background to `The GCB client needs an endpoint or set of endpoints to facilitate uploading and downloading of multiple files, usually entire projects`.

So far I have that we will not need to implement `name_contains` or `exclude_fields`. Your client will always ask for all children of a project or folder.

We will develop a new response document proposal and push to this branch to update the PR, and repeat the review.

dleehr:
Glad to see this discussion,

I commented inline on the proposal, but I have a couple other thoughts:

- For positioning/naming the endpoint, I'd suggest something like `transfer_manifest` . We're not asking for a search with fewer parameters - this would be used specifically for preparing to download/upload files.
- In DukeDSClient, the first step of a download or upload is to fetch the File/Folder paths, IDs, and checksums, which currently uses the project children endpoint. Once it has that data, it has to make a second request to get a signed URL for each file. That adds overhead to every transfer that's not part of the 15 minutes quoted for getting project children. And since those are set to expire after a few minutes, we don't fetch them until right before initiating a download.

I'll offer an over-simplified suggestion on our ideal API. 

```
GET /projects/:id/transfer_manifest
```

With a response like so:

```
[
  { "path": "/foo/bar/file.txt", 
    "md5": "866aecf23d71d084133007b6122f1be5", 
    "url": "https://swift.domain.com/v1/AUTH_xxxx/abcd/efg/?temp_url_sig=030303&temp_url_expires=1518192780&filename=file.txt" },
  { "path": "/foo/baz/file.txt", 
    "md5": "007b6122f1be5866aecf23d71d084133", 
    "url": "https://swift.domain.com/v1/AUTH_xxxx/abcd/efg/?temp_url_sig=030303&temp_url_expires=1518192780&filename=file.txt" }
]
```

dleehr:
@johnbradley feel free to tell me what's wrong with my mock response :)

johnbradley:
@dleehr I would add size to that. We use that to see if we have received the entire file.
Occasionally swift doesn't return all of the file, but doesn't error. 

```
[
  { "path": "/foo/bar/file.txt", 
    "size": "1239",
    "md5": "866aecf23d71d084133007b6122f1be5", 
    "url": "https://swift.domain.com/v1/AUTH_xxxx/abcd/efg/?temp_url_sig=030303&temp_url_expires=1518192780&filename=file.txt" },
  { "path": "/foo/baz/file.txt", 
    "size": "6745",
    "md5": "007b6122f1be5866aecf23d71d084133", 
    "url": "https://swift.domain.com/v1/AUTH_xxxx/abcd/efg/?temp_url_sig=030303&temp_url_expires=1518192780&filename=file.txt" }
]
```

dmlond:
We can provide a flat structure like what @dleehr suggests. This would most likely perform better than the children endpoint since it only deals with files and not folders.

We have a couple thoughts:

- You will most likely want the file_id with each file. If you ever find that the download url goes stale, you would be able to get a new url with the id.
- Generating a file path is equivalent to getting the 'ancestors' in the payload and constructing the path from the ancestors. We would prefer to return an array of ancestor_names, instead of a generated, unix style path, as this would work for other file system platforms.(ex. Windows).  Also folder and project names can have slashes in them and we would prefer to let clients escape these as needed for their platform. We will ensure that the array of ancestors is ordered such that joining the ancestors would return the path.
- Another thing to consider is that the system supports multiple hashing algorithms: sha1, sha256, md5, etc. So, reporting all hash algorithms/values associated with a file seems like a more flexible solution.  With your proposed design, if an md5 isn't reported this would return a null response..
- Would this endpoint need anything to support the cases where the client is not uploading/downloading an entire project?

johnbradley:
> This would most likely perform better than the children endpoint since it only deals with files and not folders.

What about the case where a user downloads a project containing an empty folder? I think it would be confusing if we do not download empty folders.

> Generating a file path is equivalent to getting the 'ancestors' in the payload and constructing the path from the ancestors. We would prefer to return an array of ancestor_names, instead of a generated, unix style path, as this would work for other file system platforms.(ex. Windows). Also folder and project names can have slashes in them and we would prefer to let clients escape these as needed for their platform. We will ensure that the array of ancestors is ordered such that joining the ancestors would return the path.

I'm fine with parsing ancestors on the client side. Putting a '/' in a filename seems like a really bad idea. I don't think you can create a Linux filename containing a slash.

> Another thing to consider is that the system supports multiple hashing algorithms: sha1, sha256, md5, etc. So, reporting all hash algorithms/values associated with a file seems like a more flexible solution. With your proposed design, if an md5 isn't reported this would return a null response..

It's fine if you want to report all hash algorithms/values. I just mentioned it in case it might be more performant on your end to produce just this one value.

> Would this endpoint need anything to support the cases where the client is not uploading/downloading an entire project?

Not that I am aware of. DukeDSClient allows users to specify multiple remote path filters with `--include` and `--exclude`, but these are handled within DukeDSClient after fetching the list of files/folders.

dleehr:
Thanks for the responses!

> You will most likely want the file_id with each file. If you ever find that the download url goes stale, you would be able to get a new url with the id.

If it doesn't add noticeable time, sure. It's the primary identifier, and it seems like clients SHOULD use it, but I couldn't think of a reason my mock response needed it :).

I think we need to consider the expiring URL issue separately. Preparing for a batch transfer of thousands of files that may take several hours to download won't work very well if the URLs expire in 5 minutes. That timeframe works well for the Web UI, but not as well for a long-running batch.

I looked at the [swift documentation on Temp URLs](https://docs.openstack.org/swift/latest/api/temporary_url_middleware.html), to see if there was any other way to generate these for this use case. There is a mechanism to assign two Temp URL secrets per container, but that doesn't really scale.

So one simple thing we could do is simply extend the expiration date for large transfers transfer (24 hours? 6 hours?) That would cut down on the frequency of expired URL exceptions,

One idea I had was a _fast lane_ that avoided the temp URLs altogether. Users that need the speed (e.g. our handful of cores that move lots of data) could get accounts on the object store and register those with Data Service. Then, for transfers, Data Service could grant access to the container. Clients could authenticate directly to the object store to download or upload, without expiring URLs. It may leak too many implementation details to be worthwhile, but opens up other possibilities for other clients, import, export, compression, delivery, etc

> Would this endpoint need anything to support the cases where the client is not uploading/downloading an entire project?

I didn't want to over-specify the initial proposal, but I was thinking that such a `transfer_manifest` endpoint could take optional parameters that are applied as filters. Those could be compatible with the `--include`/`--exclude` options that john has implemented in [DukeDSClient](https://github.com/Duke-GCB/DukeDSClient/blob/master/ddsc/core/pathfilter.py), or we could find common ground on something that works in both places. 

I think it makes sense to provide those filter parameters to the service, rather than always fetching the entire list and filtering client side. Fetching the metadata for a subset of a project (e.g. just a handful of files or single folder) should take less time than fetching the metadata for the entire project. Right now it takes slightly more time.

dmann:
@johnbradley @dleehr We updated our design document to reflect what was discussed above.

johnbradley:
@dmann Do you have any thoughts on how this endpoint would handle empty folders?

dmann:
@johnbradley We’re going to focus on listing files in this endpoint, and ignore empty directories (similar to how `git` works). We can create a separate endpoint for folder structure in the future.

dmann:
@johnbradley @dleehr I've updated the api_design document to reflect some changes and fill some gaps that I felt were necessary:
- Adding permissions definition
- Returning a complete `ancestors` payload in Response Example, which can be used by clients when uploading new files to existing folders.
- Changing `url` payload in Response Example to match the more complete response provided by the `/files/:id/url` endpoint.

Take a look and let me know if anything doesn't make sense.

Donmurry (approved):
👍

PR #1135, opened 2018-02-20
DDS-1071 project files endpoint

dmann:


dmlond (approved):
👍

dmann:
@Donmurry @dmlond Here are my current points of concern:
- Since we're not returning the `ancestor` ids, nor the `parent` payload, the client does not have a way to upload new files to existing folders using this payload. I recommend returning the full `ancestor` payload, as opposed to returning the redundant `parent` payload.
- I set the permission to `download_file` because we're returning the file's `url`. The children endpoints only require `view_project`. One option that we have, if we want to allow `view_project` to use this endpoint, is to only return the file's `url` when the `download_file` permission is set. I think that may be the way to go, and I think it's probably going to be more tricky to clearly document than it will be to code.

We don't have to make a decision about this here. We can merge this into the feature branch and continue the discussion in the feature PR. I just wanted to bring this up here first.

dmlond:
I thought we decided it would be good to make a change to api_design/DDS-1061-optimized_directory_structure_endpoint.md to so that the payload includes an array of `ancestors`: [{id: ..., kind: ..., name: ...},...] instead of `ancestor_names`: [name1, name2,...].

dmann:
@dmlond We waffled back and forth on the `ancestors` payload, but since `ancestor_names` is in the document, I thought that's what we settled on.  :-P

app/api/dds/v1/files_api.rb
```diff
@@ -21,17 +21,23 @@ class FilesAPI < Grape::API
         authorize DataFile.new(project: project), :download?
         files = project.data_files.unscope(:order).order(updated_at: :desc).where(is_deleted: false)
 
-        # plain query
-        #paginate(files)
+        files_query = headers&.fetch("Project-Files-Query", nil) || "plain"
```
dmann: @dmlond We can use this `Project-Files-Query` header when we profile this endpoint, sort of like a feature flag. We should probably remove this logic once we've determined which query is optimal.

dmlond: good idea

app/api/dds/v1/files_api.rb
```diff
@@ -1,6 +1,45 @@
...
+      params do
+        use :pagination
+      end
+      get '/projects/:id/files', root: 'results', each_serializer: DataFileSummarySerializer do
+        authenticate!
+        project = hide_logically_deleted Project.find(params[:id])
+        authorize DataFile.new(project: project), :download?
+        files = project.data_files.unscope(:order).order(updated_at: :desc).where(is_deleted: false)
+
+        files_query = headers&.fetch("Project-Files-Query", nil) || "plain"
```
dmlond: I dont see "plain" in the case options below

dmlond (approved):
This looks good to me. I was a little unclear how the feature flag worked, but it makes sense now.

Donmurry (approved):
🥇

PR #1136, opened 2018-02-20
DDS-1061 optimized directory structure endpoint

dmann:
- Adding permissions definition
- Returning a complete `ancestors` payload in Response Example, which can be used by clients when uploading new files to existing folders.
- Changing `url` payload in Response Example to match the more complete response provided by the `/files/:id/url` endpoint.


dmlond (approved):
🥇

PR #1151, opened 2018-03-22
DDS-1083 unique project slugs

dmann:


johnbradley:
This sounds fine to me. I like that the slug field is optional so no DukeDSClient changes will be necessary for this to go out. It should be pretty simple to add support for specifying the slug in DukeDSClient once it is ready.

dmlond (approved):
💯

PR #1152, opened 2018-03-27
DDS-829 design change deletion

dmlond:
This design proposes that restorable items (folders, files) should moved into the root of the project trash bin when they are deleted. This makes the project trash bin behave more consistently with the way that OS trashbins work.


dmann (approved):
👍

Donmurry (approved):
👍

johnbradley:
For DukeDSClient we only delete the entire project so this change will not have an effect since the project level delete doesn't use the trashbin. 👍

PR #1154, opened 2018-04-02
DDS-1095-purge_restore_authorization

dmlond:
In response to code review

### DIFF
--- a/spec/requests/trashbin_api_spec.rb
+++ b/spec/requests/trashbin_api_spec.rb
@@ -64,7 +64,6 @@ describe DDS::V1::TrashbinAPI do
 
   describe 'PUT /trashbin/{object_kind}/{object_id}/restore' do
     let(:url) { "/api/v1/trashbin/#{resource_kind}/#{resource_id}/restore" }
-    subject { put(url, params: payload.to_json, headers: headers) }
     let(:called_action) { 'PUT' }
     let(:parent_kind) { parent_folder.kind }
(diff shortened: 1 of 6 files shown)


spec/requests/trashbin_api_spec.rb
```diff
@@ -74,73 +73,137 @@
...
     let(:resource_id) { trashed_resource.id }
 
-    let(:payload) {{
-      parent: {
-        kind: parent_kind,
-        id: parent_id
-      }
-    }}
+    context 'without payload' do
+      subject { put(url, headers: headers) }
```
dmann, line 77: Would it work to wrap all these contexts with `it_behaves_like 'a PUT request'`?

spec/requests/trashbin_api_spec.rb
```diff
@@ -74,73 +73,137 @@
...
-    end
+        it_behaves_like 'an identified resource' do
+          let(:resource_id) { untrashed_resource.id }
+        end
 
-    it_behaves_like 'an identified resource' do
-      let(:parent_id) { "doesNotExist" }
-      let(:resource_class) { Folder }
-    end
+        it_behaves_like 'a client error' do
```
dmann, line 88: This test would read a bit better if it were wrapped in `context 'with deleted project'`, with the before block nested in the context instead of the shared example.

dmann (commented):
🔢

dmlond:
closing this for further discussion on implementing moves to different project

PR #1155, opened 2018-04-02
DDS-1090-trashbin_children_sorting

dmlond:
ensures results from trashbin/{folder|project}/{id}/chidren are sorted by updated_at ascending

### DIFF
--- a/spec/support/requests_shared_examples.rb
+++ b/spec/support/requests_shared_examples.rb
@@ -224,17 +224,26 @@ shared_examples 'a sorted index resource' do |expected_last_item_sym|
     expect(response_json).to have_key('results')
     returned_results = response_json['results']
     expect(returned_results).to be_a(Array)
+    expect(returned_results).not_to be_empty
     last_date = nil
     returned_results.each do |this_result|
       this_result_object = KindnessFactory.by_kind(this_result["kind"]).find(this_result["id"])
-      if last_date.nil?
-        last_date = this_result_object.send(sort_column)
-      elsif sort_order == "asc"
+      last_date = this_result_object.send(sort_column) if last_date.nil?
+
+      if sort_order == "asc"
         expect(this_result_object.send(sort_column)).to be >= last_date
       else
         expect(this_result_object.send(sort_column)).to be <= last_date
       end
+      last_date = this_result_object.send(sort_column)
     end
+  end
+
+  it 'should end with the expected last item' do
+    is_expected.to eq(expected_response_status)
+    response_json = JSON.parse(response.body)
+    expect(response_json).to have_key('results')
+    returned_results = response_json['results']
     expect(returned_results.last["id"]).to eq(expected_last_item.id)
   end
 end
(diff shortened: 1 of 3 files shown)


spec/support/requests_shared_examples.rb
```diff
@@ -213,6 +213,32 @@
...
+  let(:sort_order) { "desc" }
+
+  it 'should return a sorted index' do
+    is_expected.to eq(expected_response_status)
+    response_json = JSON.parse(response.body)
+    expect(response_json).to have_key('results')
+    returned_results = response_json['results']
+    expect(returned_results).to be_a(Array)
+    last_date = nil
+    returned_results.each do |this_result|
```
dmann: I like what you're doing in this test with the looping over the results and testing the `sort_column`.

spec/support/requests_shared_examples.rb
```diff
@@ -213,6 +213,32 @@
...
+      this_result_object = KindnessFactory.by_kind(this_result["kind"]).find(this_result["id"])
+      if last_date.nil?
+        last_date = this_result_object.send(sort_column)
```
dmann: It should probably be setting `last_date` each time it loops through.

spec/support/requests_shared_examples.rb
```diff
@@ -213,6 +213,32 @@
...
+  let(:expected_last_item) { send(expected_last_item_sym) }
+  let(:sort_column) { :created_at }
+  let(:sort_order) { "desc" }
+
+  it 'should return a sorted index' do
+    is_expected.to eq(expected_response_status)
+    response_json = JSON.parse(response.body)
+    expect(response_json).to have_key('results')
+    returned_results = response_json['results']
+    expect(returned_results).to be_a(Array)
```
dmann, line 226: It would be good to test that `returned_results` has more than one element.

spec/support/requests_shared_examples.rb
```diff
@@ -213,6 +213,32 @@
...
+      this_result_object = KindnessFactory.by_kind(this_result["kind"]).find(this_result["id"])
+      if last_date.nil?
+        last_date = this_result_object.send(sort_column)
+      elsif sort_order == "asc"
+        expect(this_result_object.send(sort_column)).to be >= last_date
+      else
+        expect(this_result_object.send(sort_column)).to be <= last_date
+      end
+    end
+    expect(returned_results.last["id"]).to eq(expected_last_item.id)
```
dmann: I don't think that you need this `expected_last_item` test. If you want to keep it though, it would be better suited in a separate `it` block.

dmann (changes requested):
Looks good for the most part.

spec/support/requests_shared_examples.rb
dmlond: I'll put it in its own it block. I like it because it forces the test writer to do something specific to an item to ensure it is not part of a `let (:items) { FactoryBot.create_list(:foo, 3) }` group which might have very similar created_at/updated_at entries.

spec/support/requests_shared_examples.rb
dmlond: done

dmann: Almost. I'll rephrase. Ensure `returned_results` has two or more elements. It's not useful to test that an array with a single element is sorted properly...

dmann (approved):
🔢

PR #1156, opened 2018-04-03
DDS-1098-container_deletion_parent

dmlond:
When a Folder or DataFile is deleted, it is moved to the root of the project trashbin, but its original parent is preserved for future restores.

### DIFF
--- a/app/models/container.rb
+++ b/app/models/container.rb
@@ -8,10 +8,12 @@ class Container < ActiveRecord::Base
   audited
   belongs_to :project
 	belongs_to :parent, class_name: "Folder"
+  belongs_to :deleted_from_parent, class_name: "Folder"
   has_many :project_permissions, through: :project
   has_many :tags, as: :taggable
 
   define_model_callbacks :set_parent_attribute
+
   validates :name, presence: true, unless: :is_deleted
 
   def ancestors
@@ -36,5 +38,15 @@ class Container < ActiveRecord::Base
 
   def set_project_to_parent_project
     self.project = self.parent.project if self.parent
+  end
+
+  def set_deleted_from_parent
+    if is_deleted?
+      self.deleted_from_parent_id = self.parent_id
+      self.parent_id = nil
+    elsif is_deleted_was && !is_deleted?
+      self.parent_id = self.deleted_from_parent_id
+      self.deleted_from_parent_id = nil
+    end
   end
 end
--- a/app/models/container.rb
+++ b/app/models/container.rb
@@ -54,8 +54,12 @@ class Container < ActiveRecord::Base
     elsif new_parent.is_a? Folder
       self.deleted_from_parent_id = nil
       self.parent_id = new_parent.id
-    else
+      self.project_id = new_parent.project_id
+    elsif new_parent.is_a? Project
       self.deleted_from_parent_id = nil
+      self.project_id = new_parent.id
+    else
+      raise IncompatibleParentException.new("Objects can only be restored to a dds-folder or dds-project.::Perhaps you mistyped the object_kind.")
     end
   end
 end
--- a/app/models/container.rb
+++ b/app/models/container.rb
@@ -40,12 +40,21 @@ class Container < ActiveRecord::Base
     self.project = self.parent.project if self.parent
   end
 
-  def set_deleted_from_parent
-    if is_deleted?
-      self.deleted_from_parent_id = self.parent_id
-      self.parent_id = nil
-    elsif is_deleted_was && !is_deleted?
+  def move_to_trashbin
+    self.is_deleted = true
+    self.deleted_from_parent_id = self.parent_id
+    self.parent_id = nil
+  end
+
+  def restore_from_trashbin(new_parent=nil)
+    self.is_deleted = false
+    if new_parent.nil?
       self.parent_id = self.deleted_from_parent_id
+      self.deleted_from_parent_id = nil
+    elsif new_parent.is_a? Folder
+      self.deleted_from_parent_id = nil
+      self.parent_id = new_parent.id
+    else
       self.deleted_from_parent_id = nil
     end
   end
--- a/app/models/container.rb
+++ b/app/models/container.rb
@@ -54,8 +54,8 @@ class Container < ActiveRecord::Base
     elsif new_parent.is_a? Folder
       self.deleted_from_parent_id = nil
       self.parent_id = new_parent.id
-      self.project_id = new_parent.project_id
     elsif new_parent.is_a? Project
+      self.parent_id = nil
       self.deleted_from_parent_id = nil
       self.project_id = new_parent.id
     else
--- a/spec/models/data_file_spec.rb
+++ b/spec/models/data_file_spec.rb
@@ -292,7 +292,7 @@ RSpec.describe DataFile, type: :model do
       let(:original_parent) { subject.deleted_from_parent }
       it {
         subject.move_to_trashbin
-        subject.save
+        expect(subject.save).to be_truthy
 
         expect(subject.is_deleted?).to be_truthy
         expect(subject.parent_id).to be_nil
@@ -316,7 +316,7 @@ RSpec.describe DataFile, type: :model do
         let(:original_parent) { subject.deleted_from_parent }
         it {
           subject.move_to_trashbin
-          subject.save
+          expect(subject.save).to be_truthy
 
           expect(subject.is_deleted?).to be_truthy
           expect(subject.parent_id).to be_nil
@@ -342,7 +342,7 @@ RSpec.describe DataFile, type: :model do
         let(:original_parent) { subject.deleted_from_parent }
         it {
           subject.move_to_trashbin
-          subject.save
+          expect(subject.save).to be_truthy
 
           expect(subject.is_deleted?).to be_truthy
           expect(subject.parent_id).to be_nil
@@ -368,7 +368,7 @@ RSpec.describe DataFile, type: :model do
       let(:target_project) { project }
       it {
         subject.move_to_trashbin
-        subject.save
+        expect(subject.save).to be_truthy
 
         expect(subject.is_deleted?).to be_truthy
         expect(subject.parent_id).to be_nil
@@ -392,7 +392,7 @@ RSpec.describe DataFile, type: :model do
       let(:target_project) { other_project }
       it {
         subject.move_to_trashbin
-        subject.save
+        expect(subject.save).to be_truthy
 
         expect(subject.is_deleted?).to be_truthy
         expect(subject.parent_id).to be_nil
@@ -416,7 +416,7 @@ RSpec.describe DataFile, type: :model do
       let(:new_parent) { file_versions.first }
       it {
         subject.move_to_trashbin
-        subject.save
+        expect(subject.save).to be_truthy
 
         expect {
           subject.restore_from_trashbin new_parent
--- a/spec/models/data_file_spec.rb
+++ b/spec/models/data_file_spec.rb
@@ -24,6 +24,7 @@ RSpec.describe DataFile, type: :model do
   describe 'associations' do
     it { is_expected.to belong_to(:project) }
     it { is_expected.to belong_to(:parent) }
+    it { is_expected.to belong_to(:deleted_from_parent) }
     it { is_expected.to have_many(:project_permissions).through(:project) }
     it { is_expected.to have_many(:file_versions).order('version_number ASC').autosave(true) }
     it { is_expected.to have_many(:tags) }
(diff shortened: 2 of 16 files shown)


app/models/container.rb
```diff
@@ -37,4 +39,14 @@ def parent_id=(val)
   def set_project_to_parent_project
     self.project = self.parent.project if self.parent
   end
+
+  def set_deleted_from_parent
```
dmann: I think it would be good to break this method into two methods. Something like `delete_to_trash` and `restore_to_parent`. They could encapsulate setting `is_deleted` as well as `deleted_from_parent_id` and `parent_id`. This should result in code that is easier to both read and test. 🤞

dmann (changes requested):
🍳

spec/models/data_file_spec.rb
```diff
@@ -262,53 +262,100 @@
       end
     end
 
-    describe '#set_deleted_from_parent' do
-      it { is_expected.to respond_to(:set_deleted_from_parent) }
+    describe '.move_to_trashbin' do
```
dmann: `s/\./#/`
It's an instance method.

spec/models/data_file_spec.rb
```diff
@@ -262,53 +262,100 @@
...
+        expect(subject.parent_id).to be_nil
+        expect(subject.parent).to be_nil
+        expect(subject.deleted_from_parent_id).not_to be_nil
+        expect(subject.deleted_from_parent).not_to be_nil
+        expect(subject.deleted_from_parent).to eq(original_parent)
+      }
+    end
+  end
+
+  describe '.restore_from_trashbin' do
```
dmann: `s/\./#/`
It's an instance method.

spec/models/data_file_spec.rb
```diff
@@ -262,53 +262,100 @@
...
-          expect(subject.deleted_from_parent_id).to be_nil
-          expect(subject.deleted_from_parent).to be_nil
-          expect(subject.parent).to eq(original_parent)
-        }
-      end
+      subject { child_file }
+      it {
+        expect(subject.parent_id).not_to be_nil
+        expect(subject.parent).not_to be_nil
+        original_parent = subject.parent
```
dmann: You could set `original_parent` as a let variable and then, instead of having `expect(subject.parent).not_to be_nil`, use `expect(original_parent).not_to be_nil`

spec/models/data_file_spec.rb
```diff
@@ -262,53 +262,100 @@
...
+    end
+  end
+
+  describe '.restore_from_trashbin' do
+    it { is_expected.to respond_to(:restore_from_trashbin) }
+
+    context 'to original parent' do
+      it {
+        subject.move_to_trashbin
+        subject.save
```
dmann: Make sure the call to save is truthy.

spec/models/data_file_spec.rb
```diff
@@ -262,53 +262,100 @@
...
+
+        expect(subject.is_deleted?).to be_truthy
+        expect(subject.parent_id).to be_nil
+        expect(subject.parent).to be_nil
+        expect(subject.deleted_from_parent_id).not_to be_nil
+        expect(subject.deleted_from_parent).not_to be_nil
+        original_parent = subject.deleted_from_parent
```
dmann: See above advice regarding `original_parent`

spec/models/data_file_spec.rb
```diff
@@ -262,53 +262,100 @@
...
+        expect(subject.deleted_from_parent).to be_nil
+        expect(subject.parent).to eq(original_parent)
+      }
+    end
+
+    context 'to new parent folder' do
+      let(:new_parent) { other_folder }
+      it {
+        subject.move_to_trashbin
+        subject.save
```
dmann: Make sure call to save is truthy

spec/models/data_file_spec.rb
```diff
@@ -262,53 +262,100 @@
...
+
+        expect(subject.is_deleted?).to be_truthy
+        expect(subject.parent_id).to be_nil
+        expect(subject.parent).to be_nil
+        expect(subject.deleted_from_parent_id).not_to be_nil
+        expect(subject.deleted_from_parent).not_to be_nil
+        original_parent = subject.deleted_from_parent
```
dmann: See above advice regarding `original_parent`...

dmann: Actually, if above 8 lines are the same for all these tests, then they belong in a before(:each) block as well. Just create a wrapping context called 'with trashed subject' or something.

spec/models/data_file_spec.rb
```diff
@@ -262,53 +262,100 @@
...
+        expect(subject.parent_id).not_to be_nil
+        expect(subject.parent).not_to be_nil
+        expect(subject.deleted_from_parent_id).to be_nil
+        expect(subject.deleted_from_parent).to be_nil
+        expect(subject.parent).not_to eq(original_parent)
+        expect(subject.parent).to eq(new_parent)
+      }
+    end
+
+    context 'to project root' do
```
dmann: This tests how the method behaves when passed the subject's project. What happens with a different project?

spec/models/folder_spec.rb
```diff
@@ -197,10 +197,11 @@
   end
 
 
-  describe '#restore' do
+  describe '.restore' do
```
dmann: `s/\./#/`
It's an instance method.

spec/models/folder_spec.rb
```diff
@@ -229,62 +230,40 @@
...
+            expect(child.is_deleted?).to be_falsey
+            expect(child.parent_id).to eq(subject.id)
+          }.not_to raise_error
+        }
       end
     end
   end
 
-  describe '#purge' do
+  describe '.purge' do
```
dmann: `s/\./#/`
It's an instance method.

spec/models/folder_spec.rb
```diff
@@ -313,26 +293,44 @@
     end
   end
 
-  describe '#set_deleted_from_parent' do
-    it { is_expected.to respond_to(:set_deleted_from_parent) }
+  describe '.move_to_trashbin' do
```
dmann: `s/\./#/`
It's an instance method.

spec/models/folder_spec.rb
```diff
@@ -313,26 +293,44 @@
...
+      expect(subject.parent).to be_nil
+      expect(subject.deleted_from_parent_id).not_to be_nil
+      expect(subject.deleted_from_parent).not_to be_nil
+      expect(subject.deleted_from_parent).to eq(original_parent)
+    }
+  end
 
-    context 'being restored' do
-      subject { FactoryBot.create(:folder, :deleted, :with_parent) }
+  describe '.restore_from_trashbin' do
```
dmann: `s/\./#/`
It's an instance method.

spec/models/project_spec.rb
```diff
@@ -372,7 +372,7 @@
       }
     end #purge_children
 
-    describe '#restore' do
+    describe '.restore' do
```
dmann: `s/\./#/`
It's an instance method.

spec/models/data_file_spec.rb
dmann: I'm also curious what happens when it's passes a folder from a different project.

spec/models/folder_spec.rb
```diff
@@ -196,10 +197,11 @@
   end
 
 
-  describe '#restore' do
+  describe '.restore' do
     context 'is_deleted? true' do
       before do
-        subject.update_columns(is_deleted: true)
+        subject.move_to_trashbin
+        subject.save
```
dmann: Make sure the call to save is truthy

spec/models/folder_spec.rb
```diff
@@ -228,62 +230,40 @@
...
     context 'when child is a Container' do
-      context 'from another project' do
-        let(:child) { other_folder }
-        before do
-          child.update_columns(is_deleted: true)
-          child.reload
-        end
+      before do
+        child.move_to_trashbin
+        child.save
```
dmann: Make sure the call to save is truthy

spec/models/folder_spec.rb
```diff
@@ -298,7 +278,8 @@
 
     context 'deleted' do
       before do
-        subject.update_columns(is_deleted: true)
+        subject.move_to_trashbin
+        subject.save
```
dmann: Make sure the call to save is truthy

spec/models/folder_spec.rb
```diff
@@ -311,4 +292,100 @@
       }
     end
   end
+
+  describe '.move_to_trashbin' do
```
dmann: `s/\./#/`
It's an instance method.

spec/models/folder_spec.rb
```diff
@@ -311,4 +292,100 @@
...
+      expect(subject.is_deleted?).to be_truthy
+      expect(subject.parent_id).to be_nil
+      expect(subject.parent).to be_nil
+      expect(subject.deleted_from_parent_id).not_to be_nil
+      expect(subject.deleted_from_parent).not_to be_nil
+      expect(subject.deleted_from_parent).to eq(original_parent)
+    }
+  end
+
+  describe '.restore_from_trashbin' do
```
dmann: `s/\./#/`
It's an instance method.

spec/models/folder_spec.rb
```diff
@@ -311,4 +292,100 @@
...
+    context 'to original parent' do
+      it {
+        subject.move_to_trashbin
+        subject.save
```
dmann: Make sure the call to save is truthy

dmann (changes requested):
I stopped reviewing about half way through. I found myself repeating my comments. I'll come back and review again tomorrow or Monday.

spec/models/data_file_spec.rb
dmlond: Sorry, I get these confused

spec/models/data_file_spec.rb
dmlond: The function does what it is told to do when a new project, or folder from a new project is passed. It sets these variables. This results in an invalid object based on the validations (project is immutable), which the trashbin_api restore method now communicates to the user.

app/models/container.rb
```diff
@@ -37,4 +39,27 @@ def parent_id=(val)
...
+
+  def restore_from_trashbin(new_parent=nil)
+    self.is_deleted = false
+    if new_parent.nil?
+      self.parent_id = self.deleted_from_parent_id
+      self.deleted_from_parent_id = nil
+    elsif new_parent.is_a? Folder
+      self.deleted_from_parent_id = nil
+      self.parent_id = new_parent.id
+      self.project_id = new_parent.project_id
```
dmann: The project automatically gets set to the parent's project via the `set_parent_attribute` callback (which calls `set_project_to_parent_project`). Should be able to get rid of this line and have your tests continue to pass.

app/models/container.rb
```diff
@@ -37,4 +39,27 @@ def parent_id=(val)
...
+    elsif new_parent.is_a? Project
+      self.deleted_from_parent_id = nil
+      self.project_id = new_parent.id
```
dmann: Should `self.parent` be set to `nil` here as well?

spec/models/data_file_spec.rb
```diff
@@ -260,6 +262,138 @@
...
+    it { is_expected.to respond_to(:restore_from_trashbin) }
+
+    before do
+      expect(subject.is_deleted?).to be_truthy
+      expect(subject.parent_id).to be_nil
+      expect(subject.parent).to be_nil
+      expect(original_parent).not_to be_nil
+    end
+
+    context 'to original parent' do
```
dmann, line 302: This is just one of two 'original parent' scenarios, where the original parent was a folder. The second is when the original parent was the project. I don't think it'll cause you to change the code, but it would be good to have both for test coverage.

spec/models/data_file_spec.rb
```diff
@@ -260,6 +262,138 @@
...
+          expect(subject.deleted_from_parent).to be_nil
+          expect(subject.parent).not_to eq(original_parent)
+          expect(subject.parent).to eq(new_parent)
+          expect(subject.project_id).to eq(new_parent.project_id)
+          expect(subject).not_to be_valid
+        }
+      end
+    end
+
+    context 'to original project root' do
```
dmann, line 371: This might be the second 'original parent' test I mentioned above, or at least pretty close to it. I think the key difference is to call `#restore_from_trashbin` without a target parameter.

spec/requests/trashbin_api_spec.rb
```diff
@@ -64,83 +64,178 @@
...
-      end
-    end
+          context 'that is deleted' do
+            before do
+              project.update_columns(is_deleted: true)
+            end
+            it_behaves_like 'a client error' do
+              let(:expected_response) { 404 }
+              let(:expected_reason) { "dds-project #{project.id} is permenantly deleted, and cannot restore children." }
+              let(:expected_suggestion) { "Restore to a different project." }
```
dmann: I'm not asking for a change here, but is this even possible? I'm under the impression that when a project is deleted, it purges its children. Maybe this can happen with projects/children that are currently `is_deleted` in dds?

spec/requests/trashbin_api_spec.rb
```diff
@@ -64,83 +64,178 @@
...
+        context 'resource in a folder' do
+          let(:resource) { trashed_child_resource }
+          let(:resource_kind) { trashed_child_resource.kind }
+          let(:resource_id) { trashed_child_resource.id }
+
+          before do
+            expect(parent_folder).to be_persisted
+          end
+
+          context 'that is deleted' do
```
dmann: I like this test. 👍

spec/requests/trashbin_api_spec.rb
```diff
@@ -64,83 +64,178 @@
...
+            let(:resource_kind) { 'invalid-kind' }
+          end
+
+          it_behaves_like 'a validated resource'
         end
       end
     end
 
     context 'file_version' do
+      subject { put(url, headers: headers) }
```
dmann: Why the addition of `subject { put(url, headers: headers) }` in this section?

dmann (changes requested):
A lot of good stuff. Just a few changes and questions. 🦆

spec/requests/trashbin_api_spec.rb
dmlond: It may just be defensive

dmann (approved):

PR #1157, opened 2018-04-03
No More Cross Project Restores

dmlond:
Changed to reflect that /restore will not let users restore to a different project from the original project of the deleted object.


dmann (approved):
👍

PR #1159, opened 2018-04-13
DDS-1107-trashbin_data_migration

dmlond:


lib/tasks/db_data_migrate.rake
```diff
@@ -1,3 +1,31 @@
+def purge_deleted_objects
+  if ENV["PURGE_OBJECTS"]
+    Project.where(is_deleted: true).all.each do |deleted_project|
+      deleted_project.force_purgation
```
dmann: Shouldn't this be `deleted_project.force_purgation = true`? There must be something missing in your test coverage to allow this through...  🤕

dmann (changes requested):
So close!

dmann (approved):

PR #1162, opened 2018-04-17
DDS-1109-db_data_migrate_transaction

dmlond:
ensures that deleted objects have current_transactions before they are purged


dmann (approved):
🗑

PR #1164, opened 2018-04-19
DDS-1115-upload_chunk_ignore_notfound

dmlond:
Chunk and Upload #purge_storage methods treat Not Found StorageProviderExceptions as
successful deletions

### DIFF
--- a/app/models/upload.rb
+++ b/app/models/upload.rb
@@ -115,14 +115,12 @@ class Upload < ActiveRecord::Base
 
     begin
       storage_provider.delete_object_manifest(storage_container, id)
-      self.update(purged_on: DateTime.now)
     rescue StorageProviderException => e
-      if e.message.match /Not Found/
-        self.update(purged_on: DateTime.now)
-      else
+      unless e.message.match /Not Found/
         raise e
       end
     end
+    self.update(purged_on: DateTime.now)
   end
 
   def max_size_bytes
--- a/app/models/upload.rb
+++ b/app/models/upload.rb
@@ -112,8 +112,17 @@ class Upload < ActiveRecord::Base
       chunk.purge_storage
       chunk.destroy
     end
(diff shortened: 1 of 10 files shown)


app/models/upload.rb
```diff
@@ -112,8 +112,17 @@ def purge_storage
...
     end
-    storage_provider.delete_object_manifest(storage_container, id)
-    self.update(purged_on: DateTime.now)
+
+    begin
+      storage_provider.delete_object_manifest(storage_container, id)
+      self.update(purged_on: DateTime.now)
+    rescue StorageProviderException => e
+      if e.message.match /Not Found/
+        self.update(purged_on: DateTime.now)
```
dmann: Could you put the `self.update` line after the begin/rescue block, instead of repeating it twice?

spec/models/chunk_spec.rb
```diff
@@ -113,20 +113,74 @@
...
+            expect(resp.response.code.to_i).to eq(404)
+
+            expect {
+              subject.purge_storage
+            }.not_to raise_error
+          }
+        end
+
+        context 'Other Exception' do
+          let(:fake_storage_provider) { FactoryBot.create(:storage_provider) }
```
dmann, line 137: I'm not sure how this works. The storage_provider factory is populated randomly by Faker, including the `url_root` and `auth_uri`. Unless you crafted the VCR cassette by hand, this is going to try to make calls to non-existent hosts. I'm not sure what HTTP errors are thrown due to this, but it's not the authentication failure you were looking for. Maybe try fiddling with the `service_user` or `service_pass` of the `subject.storage_provider`? You want to see an auth failure recorded in your VCR cassette.

dmann (changes requested):
A couple changes

dmann (approved):

PR #1172, opened 2018-05-03
Switching to CircleCI 2.0

dmann:


dmann:
Figured out building. Now for deployment!

PR #1193, opened 2018-06-25
DDS-1106 switch to circle 2

dmann:


.circleci/config.yml
```diff
@@ -0,0 +1,182 @@
...
+  build-deploy:
+    jobs:
+      - build
+      - deploy:
+          requires:
+            - build
+          filters:
+            branches:
+              only:
+                - circleci_test_develop
```
dmlond: what are these branches? Should they be s/circle_test_//g?

.circleci/config.yml
```diff
@@ -0,0 +1,182 @@
...
+      enabled: true
+    steps:
+      - checkout
+      - run:
+          name: Ensure heroku app name variable is set
+          command: .circleci/heroku-app-name.sh
+      - run:
+          name: Store current heroku config
+          command: |
+            heroku config -a "$(.circleci/heroku-app-name.sh)" | tail -n +2 | cut -d: -f1 | tee /tmp/heroku-config-var-names
```
dmlond: you can use `heroku config --shell` instead of `| tail -n +2`

dmlond (changes requested):

dmlond (approved):

PR #1203, opened 2018-07-09
DDS_1132_find-in-batches

dmlond:
Updates Projects, Folders, DataFiles, and Uploads with either find_in_batches, or update_all. This ensures that the runner will not run out of memory when running in production with massive numbers of these objects.

### DIFF
--- a/lib/tasks/db_data_migrate.rake
+++ b/lib/tasks/db_data_migrate.rake
@@ -149,10 +149,7 @@ end
 
 def migrate_nil_storage_container
   updated_uploads = 0
-  Upload.where(storage_container: nil).each do |u|
-    u.update_columns(storage_container: u.project_id)
-    updated_uploads += 1
-  end
+  updated_uploads = Upload.where(storage_container: nil).update_all('storage_container = project_id')
   puts "#{updated_uploads} uploads updated"
 end
 
--- a/lib/tasks/db_data_migrate.rake
+++ b/lib/tasks/db_data_migrate.rake
@@ -117,16 +117,17 @@ def migrate_nil_consistency_status
   storage_provider = StorageProvider.first
   updated_projects = 0
   updated_uploads = 0
-  projects = Project.where(is_consistent: nil).where.not(is_deleted: true)
-  puts "#{projects.count} projects with nil consistency_status."
-  projects.each do |p|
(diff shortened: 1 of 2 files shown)


lib/tasks/db_data_migrate.rake
```diff
@@ -68,24 +81,23 @@ def fill_new_authentication_service_attributes
 end
 
 def create_missing_fingerprints
-  #uploads = Upload.where.not(fingerprint_value: nil)
-  uploads = Upload.eager_load(:fingerprints).where('fingerprints.id is NULL').where.not(fingerprint_value: nil, completed_at: nil).unscope(:order)
   fingerprint_count = Fingerprint.count
   failures = []
-  puts "Creating fingerprints for #{uploads.count} uploads"
```
dmann: I don't think that you need to get rid of the `uploads` variable and the line that prints out the `uploads.count`. The variable stores an ActiveRecord relation that is lazy about making the call to retrieve objects. So, you should be able to just call `uploads.find_in_batches`. Of course, it doesn't really matter, since there shouldn't be any Uploads at this point that are missing Fingerprints.

lib/tasks/db_data_migrate.rake
```diff
@@ -105,41 +117,40 @@ def migrate_nil_consistency_status
   storage_provider = StorageProvider.first
   updated_projects = 0
   updated_uploads = 0
-  projects = Project.where(is_consistent: nil).where.not(is_deleted: true)
```
dmann: Do you hate variables?

lib/tasks/db_data_migrate.rake
```diff
@@ -105,41 +117,40 @@ def migrate_nil_consistency_status
...
+      end
+      print '.'
+      updated_projects += 1
     end
-    print '.'
-    updated_projects += 1
   end
   puts "#{updated_projects} projects updated."
 
-  uploads = Upload.where(is_consistent: nil)
```
dmann: :(

dmann (changes requested):
Try to add those variables back in to keep things DRY. If it doesn't work, just add a comment and I'll approve the PR.

dmann (approved):

PR #1216, opened 2018-07-26
DDS-1136 truncate job transactions table

dmann:


app/models/job_transaction.rb
```diff
@@ -5,4 +5,22 @@ class JobTransaction < ActiveRecord::Base
   validates :key, presence: true
   validates :request_id, presence: true
   validates :state, presence: true
+
```
dmlond: would this dry things up?
```
scope :oldest_completed, ->  {  unscope(:order).order(:created_at).where(state: 'complete') }
scope :orphans, -> { where(request_id: select(:request_id).group(:request_id).having('count(*) = 1')) }

def self.oldest_completed_at
  oldest_completed.first&.created_at
end

def self.oldest_orphan_created_at
    oldest_completed.orphans.first&.created_at
end

def self.delete_all_complete_jobs(created_before: Time.now)
    where('(request_id, key) in (?)', select(:request_id, :key).where(state: 'complete').where('created_at < ?', created_before)).delete_all
end

def self.delete_all_orphans(created_before: Time.now)
    orphans.where('created_at < ?', created_before).delete_all
end
```

dmlond (changes requested):

app/models/job_transaction.rb
dmann: The :orphans scope would help, I'll add it and push up the change shortly.

The :oldest_completed scope wouldn't help and chaining it in `.oldest_orphan_created_at` would cause that method to only find completed orphans...

dmlond (approved):
👍 awesome!

PR #1217, opened 2018-08-02
DDS-1136 truncate job transactions table

dmann:
- Changing `delete_all_complete_jobs` to remove JobTransactions(JT) matching the completed JT's `request_id` AND its `transactionable`. The previous code was deleting all JT from a request, which is more greedy than I intended.
- Orphans are JTs that aren't associated with jobs, which can be determined via from the JT `state`. A JT with `state` of 'created' or 'updated' is not a job. So, a JT with `state` of 'created' or 'updated' that doesn't have any jobs matching its `request_id` and `transactionable`, is not considered an orphan.


dmlond (approved):
👍

PR #1220, opened 2018-08-21
DDS-1141-NeedlessProjectUpdates

dmlond:
prevent project_permissions updates without any changes from updating the project etag

### DIFF
--- a/spec/support/concerns/project_updater_shared_examples.rb
+++ b/spec/support/concerns/project_updater_shared_examples.rb
@@ -13,21 +13,42 @@ shared_examples 'a ProjectUpdater' do
 
   describe '#update_project_etag' do
     let!(:original_project_etag) { project.etag }
-    it {
+    let(:audit_comment) { nil }
+
+    before do
+      subject.audit_comment = audit_comment
       expect(create_subject).to be_truthy
       expect(change_subject).to be_truthy
       subject.update_project_etag
       project.reload
       subject.reload
-      last_project_audit = project.audits.last
-      last_subject_audit = subject.audits.last
+    end
+
+    it {
       expect(project.etag).not_to eq(original_project_etag)
-
-      expected_comment = last_subject_audit.comment ? last_subject_audit.comment.merge({raised_by_audit: last_subject_audit.id}) : {raised_by_audit: last_subject_audit.id}
-      expect(last_project_audit.comment.symbolize_keys).to eq(expected_comment)
-
-      expect(last_project_audit.request_uuid).to eq(last_subject_audit.request_uuid)
     }
+
+    describe 'with audit comment' do
+      let(:last_project_audit) { project.audits.last }
+      let(:last_subject_audit) { subject.audits.last }
+
+      context 'present' do
+        let(:audit_comment) {{"foo" => "bar"}}
+        let(:expected_comment) { audit_comment.merge({raised_by_audit: last_subject_audit.id}) }
+        it {
+          expect(last_project_audit.comment.symbolize_keys).to eq(expected_comment.symbolize_keys)
+          expect(last_project_audit.request_uuid).to eq(last_subject_audit.request_uuid)
+        }
+      end
+
+      context 'absent' do
+        let(:expected_comment) { {raised_by_audit: last_subject_audit.id} }
+        it {
+          expect(last_project_audit.comment.symbolize_keys).to eq(expected_comment)
+          expect(last_project_audit.request_uuid).to eq(last_subject_audit.request_uuid)
+        }
+      end
+    end
   end
 
   context 'after create' do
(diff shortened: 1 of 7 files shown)


spec/support/concerns/project_updater_shared_examples.rb
```diff
@@ -0,0 +1,71 @@
...
+
+  context 'after update' do
+    before do
+      expect(create_subject).to be_truthy
+    end
+
+    context 'without change' do
+      it {
+        is_expected.not_to receive(:update_project_etag)
+        expect(subject.saved_changes?).to be_falsey
```
dmann: Why is it `subject.saved_changes?` in this example and `subject.changed?` in the next example?

spec/support/concerns/project_updater_shared_examples.rb
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  describe '#update_project_etag' do
+    let!(:original_project_etag) { project.etag }
+    it {
+      expect(create_subject).to be_truthy
+      expect(change_subject).to be_truthy
+      subject.update_project_etag
+      project.reload
+      subject.reload
```
dmann, line 24: Let's tease apart this example. Start by putting the above 5 lines in a `before(:each)`.

spec/support/concerns/project_updater_shared_examples.rb
```diff
@@ -0,0 +1,71 @@
...
+      last_project_audit = project.audits.last
+      last_subject_audit = subject.audits.last
```
dmann: These last audit variables can be declared using `let` (if they're needed at all)

spec/support/concerns/project_updater_shared_examples.rb
```diff
@@ -0,0 +1,71 @@
...
+      expect(project.etag).not_to eq(original_project_etag)
```
dmann: This should be its own example:
`it { expect(project.etag).not_to eq(original_project_etag) }`

spec/support/concerns/project_updater_shared_examples.rb
```diff
@@ -0,0 +1,71 @@
...
+
+      expected_comment = last_subject_audit.comment ? last_subject_audit.comment.merge({raised_by_audit: last_subject_audit.id}) : {raised_by_audit: last_subject_audit.id}
+      expect(last_project_audit.comment.symbolize_keys).to eq(expected_comment)
+
+      expect(last_project_audit.request_uuid).to eq(last_subject_audit.request_uuid)
```
dmann: This should be its own example.

spec/support/concerns/project_updater_shared_examples.rb
```diff
@@ -0,0 +1,71 @@
...
+      expect(create_subject).to be_truthy
+      expect(change_subject).to be_truthy
+      subject.update_project_etag
+      project.reload
+      subject.reload
+      last_project_audit = project.audits.last
+      last_subject_audit = subject.audits.last
+      expect(project.etag).not_to eq(original_project_etag)
+
+      expected_comment = last_subject_audit.comment ? last_subject_audit.comment.merge({raised_by_audit: last_subject_audit.id}) : {raised_by_audit: last_subject_audit.id}
```
dmann: Logic like this is "code smell". Or is it "test smell"? 😛 We can control the subject's audit comment with `subject.audit_comment = {'foo' => 'bar'}` or `subject.audit_comment = nil`. Maybe use the former as the default and then test the latter as a separate context. If you use a let variable, you can easily override it for your nil audit_comment context.

dmann (changes requested):
This looks good. My only request is to break up the example that tests the `#update_project_etag` method functionality.

spec/support/concerns/project_updater_shared_examples.rb
dmlond: thanks. I had forgotten about audit_comment

spec/support/concerns/project_updater_shared_examples.rb
```diff
@@ -13,21 +13,42 @@
...
-      last_subject_audit = subject.audits.last
+    end
+
+    it {
       expect(project.etag).not_to eq(original_project_etag)
+    }
 
-      expected_comment = last_subject_audit.comment ? last_subject_audit.comment.merge({raised_by_audit: last_subject_audit.id}) : {raised_by_audit: last_subject_audit.id}
-      expect(last_project_audit.comment.symbolize_keys).to eq(expected_comment)
+    describe 'with audit comment' do
```
dmann: This should be a `context` instead of a `describe`. If the name of the block fits the format of "with ...", then usually it should be a `context`.

dmann (changes requested):
Looking good. Just requesting a minor syntax change.

dmann (approved):
👍

PR #1224, opened 2018-08-28
DDS-1147 factory bot sequence for unique attributes

dmann:
This will hopefully reduce the number of transient CI test run failures.


dmlond (approved):
nice

PR #1226, opened 2018-09-04
DDS-1148 Set background job audit env

dmann:


spec/models/application_audit_spec.rb
```diff
@@ -194,11 +192,66 @@
...
+        it { expect(described_class.current_request_uuid).to eq nonexistent_request_uuid }
+        it { expect(described_class.current_user).to be_nil }
+        it { expect(described_class.current_remote_address).to be_nil }
+        it { expect(described_class.current_comment).to be_nil }
+      end
+
+      context 'when audit env is populated and called with nonexistent request_uuid' do
+        let(:nonexistent_request_uuid) { SecureRandom.uuid }
+        before(:each) do
+          described_class.current_request_uuid = SecureRandom.uuid
```
dmlond, line 244: it took me a minute to understand what was going on here, but I get it now. It might benefit from code comments to state that setting the current environment from a request_uuid is destructive to any environment set on the model accessors

dmann: I added a comment. Check it out and let me know what you think.

dmlond (approved):
👍

PR #1236, opened 2018-10-05
DDS-1157 job transaction orphan cleanup bug fix

dmann:
- Fixes the bug that causes job transaction orphan clean up to hang
- Add job_transaction:clean_up:all task that calls clean up tasks in the correct order. This should make job scheduling easier after deployment.


dmlond (approved):
looks good

PR #1240, opened 2018-10-08
StorageProvider is now STI

dmlond:


spec/lib/tasks/db_data_migrate_rake_spec.rb
```diff
@@ -88,6 +88,38 @@
...
+      it {
+        expect(untyped_storage_provider).not_to be_a default_type
+        expect {
+          invoke_task expected_stderr: Regexp.new("1 untyped storage_providers changed to #{default_type}")
+        }.to change{
+          StorageProvider.where(type: nil).count
+        }.by(-1)
+        expected_to_be_typed_storage_provider = StorageProvider.find(untyped_storage_provider.id)
+        expect(expected_to_be_typed_storage_provider).to be_a default_type
+        swift_storage_provider.reload
```
dmann: This should be persisted earlier in the test.

dmann (changes requested):
One change, the rest looks good though.

dmann (approved):
Looks good :shipit:

PR #1242, opened 2018-10-09
DDS-1159 tags labels timing out

dmann:
BUG FIX: Refactoring sql query for `GET /tags/labels` by object_kind


dmlond (approved):
that was JimmyJohns Fast

PR #1244, opened 2018-10-12
DDS-1155-application_default_storage_provider

dmlond:


lib/tasks/db_data_migrate.rake
```diff
@@ -208,6 +208,19 @@ def populate_nil_project_slugs
   puts " #{slug_count} Project slugs populated."
 end
 
+def set_default_storage_provider
+  if StorageProvider.any?
+    if StorageProvider.where(is_default: true).any?
+      $stderr.puts "0 storage_provider default statuses changed"
```
dmann: Why are all of these using `$stderr.puts`? Seems to me that line 214 and 217 should be normal `puts` statements.

lib/tasks/db_data_migrate.rake
```diff
@@ -208,6 +208,19 @@ def populate_nil_project_slugs
...
+    else
+      StorageProvider.first.update(is_default: true)
+      $stderr.puts "first storage_provider changed to default storage_provider"
+    end
+  else
+    $stderr.puts "no storage_providers found"
```
dmann: If there aren't any storage providers, should an exception be thrown here? Or maybe we need a separate rake task to ensure that the environment is properly setup? (The latter would be a whole separate task.)

spec/lib/tasks/db_data_migrate_rake_spec.rb
```diff
@@ -125,6 +125,48 @@
       }
     end
 
+    context 'with any storage providers' do
```
dmann: Should this read 'without any storage providers'?

spec/models/swift_storage_provider_spec.rb
```diff
@@ -308,5 +308,46 @@
...
+    end
+  end
+
+  describe '.default' do
+    let(:default_storage_provider) { FactoryBot.create(:swift_storage_provider, :default) }
+    let(:not_default_storage_provider) { FactoryBot.create(:swift_storage_provider) }
+    subject { StorageProvider.default }
+    it { expect(described_class).to respond_to(:default) }
+    it {
+      expect(default_storage_provider).to be_persisted
```
dmann: What happens when `default_storage_provider` isn't persisted?

dmann (changes requested):
Looks good. I just have a couple questions that I'd like answered/addressed before approving.

lib/tasks/db_data_migrate.rake
dmlond: when you run a task with the heroku cli, it will filter out data written to STDOUT.

lib/tasks/db_data_migrate.rake
dmlond: I think we need some better post-deployment monitoring, maybe with a properly functioning `/app/status` endpoint. This is something we could remedy without needing to rerun the entire unit-test suite

spec/lib/tasks/db_data_migrate_rake_spec.rb
dmlond: yes

spec/models/swift_storage_provider_spec.rb
dmlond: it returns nil and fails the test

dmlond: it fails the test because it tries to make the not_default_storage_provider is_default: true, then it tries to make the default_storage_provider is_default true when it finally calls that let variable, which cannot be is_default true

dmann (approved):

PR #1246, opened 2018-10-16
DDS-1160 refactor job transaction orphan cleanup

dmann:
- Refactored JobTransaction code to build more scalable SQL
- BATCH_SIZE env for job_transaction:clean_up:orphans (default: 50000)


dmlond (approved):
this looks really good.

PR #1249, opened 2018-10-30
DDS-1163-storage_provider_abstract_interface

dmlond:


spec/models/swift_storage_provider_spec.rb
```diff
@@ -7,6 +7,274 @@
...
+    let(:expected_meta) {
+      {
+      "content-length" => "#{upload.size}"
+      }
+    }
+    let(:chunk) { FactoryBot.create(:chunk, :skip_validation, upload: upload) }
+
+    it_behaves_like 'A StorageProvider'
+
+    it {
```
dmann: The `it {}` syntax should really only be used for single line tests and shouldn't contain multiple `expect`s. If you have multiple expectations in an `it` block, it should have a description and be in the do/end form.

spec/factories/chunks.rb
```diff
@@ -1,13 +1,13 @@
 FactoryBot.define do
   factory :chunk do
-    upload
+    upload { create(:upload, :skip_validation) }
```
dmann: Why is `:skip_validation` required to create an upload for a chunk?

dmann: I see that `:skip_validation` has been used to build associations in several of the factories. Instead of commenting on every instance, just assume that the above question applies to all of the updated factories.

spec/factories/uploads.rb
```diff
@@ -10,15 +10,11 @@
     is_consistent { true }
 
     trait :with_chunks do
-      chunks { [ build(:chunk, number: 1) ] }
+      chunks { [ create(:chunk, :skip_validation, number: 1) ] }
```
dmann: I don't like calling `create` on traits that add child objects. In this case, if you want call `build( :upload, :with chunks)`, the chunks will be persisted and then added to the build (but not persisted) upload. I'm /guessing/ that what was causing problems here is that, even when calling `build` on a chunk, FactoryBot would try to create an upload for it. What would probably fix that issue is to pass `upload: nil` like:
```suggestion
      chunks { [ build(:chunk, upload: nil, number: 1) ] }
```
Please let me know if I'm completely off base. Hopefully we can find a solution that makes us both happy.

spec/models/chunk_spec.rb
```diff
@@ -1,16 +1,23 @@
...
+  include_context 'mocked StorageProvider Interface'
+  let(:upload) { FactoryBot.create(:upload, :skip_validation, storage_provider: mocked_storage_provider) }
+  subject { FactoryBot.create(:chunk, :skip_validation, upload: upload) }
+  include_context 'mock Chunk StorageProvider'
 
   let(:expected_object_path) { [subject.upload_id, subject.number].join('/')}
   let(:expected_sub_path) { [subject.storage_container, expected_object_path].join('/')}
-  let(:expected_expiry) { subject.updated_at.to_i + storage_provider.signed_url_duration }
-  let(:expected_url) { storage_provider.build_signed_url(subject.http_verb, expected_sub_path, expected_expiry) }
+  let(:expected_expiry) { subject.updated_at.to_i + (60*5) }
```
dmann: Might want to use something besides `.updated_at` since expiry is based on `Time.now` now.

spec/models/chunk_spec.rb
```diff
@@ -19,7 +26,9 @@
   end
 
   describe 'validations' do
-    it { is_expected.to validate_presence_of(:upload_id) }
+    it {
+      is_expected.to validate_presence_of(:upload_id)
```
dmann: I don't understand this change. Does your editor prefer this to a single line?

spec/models/chunk_spec.rb
```diff
@@ -83,111 +98,60 @@
...
-      end
+        }.not_to raise_error
+      }
     end
   end
 
-  describe '#total_chunks' do
-    it { is_expected.to respond_to :total_chunks }
-    it { expect(subject.total_chunks).to eq(subject.upload.chunks.count ) }
+  describe 'Default StorageProvider' do
```
dmann: This `describe` should probably be a `context`. If it's a context, it would read better using a description followed the "with .."/"when .." convention.

spec/models/chunk_spec.rb
```diff
@@ -83,111 +98,60 @@
...
+    it {
```
dmann: It's not clear what the goal of this test is. Please add a description and convert it from single line to do/end format.

spec/models/swift_storage_provider_spec.rb
```diff
@@ -1,13 +1,462 @@
...
+            .and_return(expected_meta)
+          expect {
+            expect(subject.is_initialized?(project)).to be_truthy
+          }.not_to raise_error
+        end
+      end
+
+      context 'project container does not exist' do
+        it 'should return false' do
+          is_expected.to receive(:get_container_meta)
```
dmann, line 131: What is the call to `#get_container_meta` returning? Does it default to nil?

spec/models/swift_storage_provider_spec.rb
```diff
@@ -1,13 +1,462 @@
...
+        expect {
+          expect(subject.single_file_upload_url(upload)).to eq(expected_url)
+        }.not_to raise_error
+      end
+    end
+
+    describe '#initialize_chunked_upload' do
+      it 'should not do anything to initialize a chunked upload in swift' do
+        expect {
+          subject.initialize_chunked_upload(upload)
```
dmann, line 170: Does `#initialize_chunked_upload` return anything on success?

spec/models/swift_storage_provider_spec.rb
```diff
@@ -1,13 +1,462 @@
...
+          expect(chunk.upload.chunks.count).to be < subject.chunk_max_number
+          expect(subject.chunk_max_exceeded?(chunk)).to be_falsey
+        end
+      end
+
+      context 'chunk.upload.chunks.count = chunk_max_number' do
+        subject { FactoryBot.create(:swift_storage_provider, chunk_max_number: chunk.upload.chunks.count) }
+        it 'should return true' do
+          expect(chunk.upload.chunks.count).to eq(subject.chunk_max_number)
+          expect(subject.chunk_max_exceeded?(chunk)).to be_truthy
```
dmann: If the chunk count is equal to the chunk_max_number, then the chunk_max has not yet been exceeded, right?

spec/models/swift_storage_provider_spec.rb
```diff
@@ -1,13 +1,462 @@
...
+        end
+      end
+
+      context 'chunk.upload.chunks.count = chunk_max_number' do
```
dmann: The description for this context is exactly the same as the one before it. :spaghetti:

spec/models/swift_storage_provider_spec.rb
```diff
@@ -1,13 +1,462 @@
...
+
+      context 'size mismatch' do
+        let(:expected_meta) {
+          {
+          "content-length" => "#{upload.size - 10}"
+          }
+        }
+
+        it 'should raise an IntegrityException' do
+          is_expected.to receive(:put_object_manifest)
```
dmann, line 255: What does `#put_object_manifest` normally return?

dmann (changes requested):
I've only reviewed the specs through `spec/models/swift_storage_provider_spec.rb` so far. I'll try to finish reviewing tomorrow morning. We really need to something to help keep PRs from getting this big, maybe a courtesy bot?

spec/factories/chunks.rb
dmlond: Whenever tests use allow(object).to receive(:storage_provider).and_return(mocked_storage_provider) (or expect..) on an Upload or Chunk, the validations will still fail because the object is being created before the mock is applied. This is especially challenging for objects that create (objects that create) uploads.

spec/factories/uploads.rb
dmlond: I think this is a problem with the way we use :with_chunks.  Most of our test create(:upload, :with_chunks). If I pass upload: nil then all of these creates will fail because the resulting upload.chunks would be invalid (they need an upload). The few places were we build an upload, we should probably just build chunks explicitly.

spec/models/chunk_spec.rb
dmlond: good catch

spec/models/chunk_spec.rb
dmlond: good catch. I moved the line down to add binding.pry and forgot to move it back after I removed the binding.pry

spec/models/chunk_spec.rb
dmlond: Good idea

spec/models/chunk_spec.rb
dmlond: will do

spec/models/swift_storage_provider_spec.rb
dmlond: This method returns (implied nil) if the response is 404.

spec/models/swift_storage_provider_spec.rb
dmlond: it does not do anything at all (it just returns) in the swift implementation since it isn't required. I can have it return true if you want. Then when we implement the S3 implementation it will need to return true or raise an exception.

spec/models/swift_storage_provider_spec.rb
dmlond: The context of when this call is being made is when a new chunk is being attempted at the moment when number of stored chunks on the upload is equal to the max. This new chunk should be invalid as it would exceed the chunk_max_number.

spec/models/swift_storage_provider_spec.rb
dmlond: good catch

spec/models/swift_storage_provider_spec.rb
dmlond: nil

spec/models/swift_storage_provider_spec.rb
dmann: Maybe change the method name from `chunk_max_exceeded?` to something like `chunk_max_met?` or `allow_another_chunk?` or `just_one_more_chunk_its_wafer_thin`?

spec/models/upload_spec.rb
```diff
@@ -219,256 +239,152 @@
...
-      it { expect(subject.minimum_chunk_size).to eq(expected_minimum_chunk_size) }
-    end
+    it {
+      expect(subject.minimum_chunk_size).to eq(mocked_storage_provider.suggested_minimum_chunk_size(subject))
+    }
   end
 
-  describe 'swift methods', :vcr do
-    subject { FactoryBot.create(:upload, :swift, :with_chunks) }
+  describe 'StorageProvider Methods' do
```
dmann: Is this describe block needed? When it was 'swift methods', I think the purpose was to enable VCR for the block. But now, it just feels out of place.

spec/models/upload_spec.rb
```diff
@@ -219,256 +239,152 @@
...
-          object = [subject.id, chunk.number].join('/')
-          subject.storage_provider.delete_object(subject.storage_container, object)
+    describe '#complete_and_validate_integrity' do
+      subject { FactoryBot.create(:upload, :skip_validation, :with_chunks, is_consistent: false, storage_provider: mocked_storage_provider) }
+
+      it { is_expected.to respond_to :complete_and_validate_integrity }
+
+      context 'with valid reported size and chunk hashes' do
+        it 'should set is_consistent to true, leave error_at and error_message null' do
+          expect(mocked_storage_provider).to receive(:complete_chunked_upload)
```
dmann: This is another method that returns 'nil' on success?

spec/models/upload_spec.rb
```diff
@@ -219,256 +239,152 @@
...
-              original_chunks.each do |chunk|
-                resp = HTTParty.head(
-                  "#{subject.storage_provider.storage_url}/#{chunk.sub_path}",
-                  headers: subject.storage_provider.auth_header
-                )
-                expect(resp.response.code.to_i).to eq(404)
-              end
-            }
+      context 'StorageProviderException' do
+        it {
```
dmann: Please change this to an `it` do/end block with a description.

spec/models/upload_spec.rb
```diff
@@ -219,256 +239,152 @@
...
+          subject.chunks.each do |chunk|
```
dmann: It's a good idea to check that `subject.chunks` is not empty before doing a loop like this.

spec/models/upload_spec.rb
```diff
@@ -219,256 +239,152 @@
...
+
+          expect {
+            expect {
+              subject.purge_storage
+            }.to change{Chunk.count}.by(-original_chunks_count)
+          }.to raise_error(unexpected_exception)
+
+          subject.reload
+          expect(subject.purged_on).to be_nil
+          expect(subject.purged_on).to be_nil
```
dmann: Accidentally checking twice that `subject.purged_on` is nil.

spec/models/upload_spec.rb
```diff
@@ -219,256 +239,152 @@
...
+          expect(subject.chunks.count).to eq 0
+        }
+      end
+
+      context 'no StorageProviderException' do
+        it 'should purge successfully' do
+          subject.chunks.each do |chunk|
```
dmann: Would be good to check that `subject.chunks` isn't empty.

spec/models/upload_spec.rb
```diff
@@ -219,256 +239,152 @@
...
-              )
-              expect(resp.response.code.to_i).to eq(404)
-            end
-          }
+            }.to change{Chunk.count}.by(-original_chunks_count)
+          }.not_to raise_error
+
+          subject.reload
+          expect(subject.purged_on).not_to be_nil
+          expect(subject.purged_on).to be >= purge_time
```
dmann: If you used the TimeHelpers to freeze time, the above two lines could be replaced with `expect(subject.purged_on).to eq purge_time`

To use the TimeHelpers, add the following inside `Rspec.describe Upload..` block:
```
include ActiveSupport::Testing::TimeHelpers
```
And then just add the following to whatever context needs time frozen:
```
  around(:each) do |example|
    travel_to(Time.now) do #freeze_time
      example.run
    end
  end
```

spec/models/upload_spec.rb
```diff
@@ -219,256 +239,152 @@
...
+        expect(upload).to be_valid
+      end
+    end
+
+    context 'completed upload' do
+      it 'should be valid when created with a swift_storage_provider' do
+        StorageProvider.delete_all
+        default_storage_provider = FactoryBot.create(:swift_storage_provider, :default)
+        expect(StorageProvider.default).not_to be_nil
+        upload = FactoryBot.create(:upload, storage_provider: StorageProvider.default)
```
dmann: The above four lines could be moved to a before block, outside of the current context because it's used in both contexts inside 'Default StorageProvider':
```suggestion
      before(:example) do
        StorageProvider.delete_all
        default_storage_provider = FactoryBot.create(:swift_storage_provider, :default)
        expect(StorageProvider.default).not_to be_nil
      end
      let(:upload) { FactoryBot.create(:upload, storage_provider: StorageProvider.default) }
```
or even:
```suggestion
      before(:example) do
        StorageProvider.delete_all
        expect(default_storage_provider).to be_persisted
        expect(StorageProvider.default).to eq default_storage_provider
      end
      let(:default_storage_provider) { FactoryBot.create(:swift_storage_provider, :default) }
      let(:upload) { FactoryBot.create(:upload, storage_provider: StorageProvider.default) }
```

spec/requests/projects_api_spec.rb
```diff
@@ -345,37 +345,6 @@
...
-              expect(root_file.is_deleted?).to be_falsey
-              is_expected.to eq(204)
-              expect(root_folder.reload).to be_truthy
-              expect(root_folder.is_deleted?).to be_truthy
-              expect(root_file.reload).to be_truthy
-              expect(root_file.is_deleted?).to be_truthy
-            }
-          end
-
-          context 'with queued ActiveJob' do
```
dmann, line 370: I understand removing the 'with inline ActiveJob' context, but don't you think it's useful to ensure that the `ChildPurgationJob` is enqueued?

spec/serializers/upload_preview_serializer_spec.rb
```diff
@@ -16,16 +19,17 @@
...
     end
   end
 
   context 'when upload has error' do
-    let(:resource) { FactoryBot.create(:upload, :with_chunks, :with_error) }
+    let(:resource) { FactoryBot.create(:upload, :with_chunks, :with_error, :skip_validation, storage_provider: mocked_storage_provider) }
     it_behaves_like 'a json serializer' do
-      it { is_expected.to include(expected_attributes) }
+      it {
+        is_expected.to include(expected_attributes) }
```
dmann: :point_up:

spec/support/concerns/storage_provider_shared_examples.rb
```diff
@@ -0,0 +1,88 @@
...
+
+  before do
+    allow_any_instance_of(Upload).to receive(:storage_provider)
+      .and_return(mocked_storage_provider)
+  end
+end
+
+shared_context 'mock Chunk StorageProvider' do |on: []|
+  let(:targets) {
+    if on.empty?
```
dmann, line 43: Might be good to `expect(on).to be_an Array` before using it

spec/support/concerns/storage_provider_shared_examples.rb
```diff
@@ -0,0 +1,88 @@
...
+end
+
+shared_examples 'A StorageProvider' do
+  it { is_expected.to be_a StorageProvider }
+
+  it { is_expected.to respond_to(:signed_url_duration) }
+  it { expect(subject.signed_url_duration).to eq 60*5 } # 5 minutes
+
+  it { is_expected.to respond_to(:expiry) }
+  it { expect(subject.expiry).to eq Time.now.to_i + subject.signed_url_duration }
```
dmann: Should probably freeze time here to prevent transient test failures.

dmann (changes requested):
I finished reviewing the specs!

spec/models/upload_spec.rb
dmlond: yes

spec/models/upload_spec.rb
dmlond: sure

spec/models/upload_spec.rb
dmlond: ok

spec/models/upload_spec.rb
dmlond: good catch

spec/models/upload_spec.rb
dmlond: ok

spec/models/swift_storage_provider_spec.rb
dmlond: how about `chunk_max_reached?`

spec/serializers/upload_preview_serializer_spec.rb
dmlond: unpryed

spec/models/upload_spec.rb
dmlond: done

spec/models/upload_spec.rb
dmlond: evidently not, as long as I remove the before block that mocks :max_size_bytes as well (which wasnt being used)

spec/models/upload_spec.rb
dmlond: good idea. I improved on it a bit.

spec/support/concerns/storage_provider_shared_examples.rb
dmlond: done

spec/support/concerns/storage_provider_shared_examples.rb
dmlond: good idea

spec/models/swift_storage_provider_spec.rb
dmann: Works for me!

lib/tasks/storage_provider.rake
```diff
@@ -1,155 +1,57 @@
...
-          end
-        rescue StorageProviderException => e
-          puts "container #{container} object #{object} returned by get_container_objects, but metadata not accessible? #{e.message}"
-        end
-      end
+namespace :storage_provider do
+  desc "creates a storage_provider using ENV"
+  task create: :environment do
+    unless ENV['STORAGE_PROVIDER_TYPE']
+      $stderr.puts "YOU MUST SET ENV[STORAGE_PROVIDER_TYPE]"
```
dmann: I think it would be useful to return the list of supported storage provider types in this error message.

app/models/chunk.rb
```diff
@@ -44,31 +44,21 @@ def sub_path
   end
 
   def expiry
```
dmann: Since the StorageProvider now handles building the signed url, is `Chunk#expiry` still being used anywhere?

spec/models/swift_storage_provider_spec.rb
dmlond: done

spec/factories/chunks.rb
dmlond: I was able to remove the use of :skip_validation from the fingerprints and chunks factory upload creation. I also removed :skip_validation from many tests.

spec/factories/uploads.rb
dmlond: fixed

app/models/storage_provider.rb
```diff
@@ -13,61 +13,72 @@ def self.default
...
   end
 
-  def delete_object_manifest(container, object)
-    resp = HTTParty.delete(
-      "#{storage_url}/#{container}/#{object}?multipart-manifest=delete",
-      headers: auth_header
-    )
-    ([200,204].include?(resp.response.code.to_i)) ||
-      raise(StorageProviderException, resp.body)
+  def single_file_upload_url(upload)
```
dmann, line 41: I like that you added this method. Looking forward to updating our api to support this.

app/models/upload.rb
```diff
@@ -80,22 +78,12 @@ def has_integrity_exception?
...
   end
 
-  def create_and_validate_storage_manifest
-    begin
-      response = storage_provider.put_object_manifest(storage_container, id, manifest, content_type, name)
-      meta = storage_provider.get_object_metadata(storage_container, id)
-      unless meta["content-length"].to_i == size
-        raise IntegrityException, "reported size does not match size computed by StorageProvider"
-      end
+  def complete_and_validate_integrity
```
dmann, line 81: I like the new name! :100:

dmann (changes requested):
Code looks good. Just a couple comments.

app/models/chunk.rb
dmlond: probably not :)

lib/tasks/storage_provider.rake
dmlond: 🤔 I thought it did. I will make sure it does

spec/requests/projects_api_spec.rb
dmlond: good idea. I added it back

app/models/chunk.rb
dmlond: removed

dmann (approved):
Looks good :shipit:

PR #1253, opened 2018-11-15
DDS-1165 ldap identity provider affiliates pagination

dmann:
This is an attempt to optimize the retrieval of affiliates from LDAP. Instead of instantiating a `User` for every entry return by LDAP, `LdapIdentityProvider#affiliates` will only instantiate those within the range of `#affiliates_offset` and `#affiliates_limit`. The rest of the array will be filled with `nil`. Hopefully this will use less memory and make for faster response times.

To turn this feature on, set the environment variable FF_LDAP_PAGINATION to 'ON'.


app/api/dds/v1/auth_providers_api.rb
```diff
@@ -84,6 +84,11 @@ def unsupported_affiliate_search_error!
         affiliate_params = declared(params, {include_missing: false}, [:full_name_contains, :username, :email])
         auth_service = AuthenticationService.find(params[:id])
         unsupported_affiliate_search_error! unless auth_service.identity_provider
+        if ENV['FF_LDAP_PAGINATION']
+          p = paginate(Kaminari.paginate_array(Array.new(500)))
```
dmlond, line 88: what is this magic number 500? Can it be set programmatically from pagination request headers, e.g. per-page?

app/models/ldap_identity_provider.rb
```diff
@@ -56,8 +58,12 @@ def ldap_search(filter:)
       size: 500,
```
dmlond, line 58: ahh, is this same 500? we should maybe put this in config.

app/models/ldap_identity_provider.rb
```diff
@@ -56,8 +58,12 @@ def ldap_search(filter:)
...
       return_results: false
     ) do |entry|
-      if new_user = ldap_entry_to_user(entry)
-        results << new_user
+      if valid_ldap_entry? entry
+        if results.length < affiliates_offset.to_i || affiliates_limit && results.length >= affiliates_limit + affiliates_offset.to_i
```
dmlond, line 62: could you do something like
```
results = Array.new(500)
iter = 0
success = ldap_conn.search(..) do |entry|
  if valid_ldap_entry? entry
    limit = affiliates_limit ? affiliates_limit + affiliates_offset.to_i : results.length
    if iter >= affiliates_limit && iter <= limit
     results << ldap_entry_to_user(entry)
    end
  end
  iter = iter + 1
end
```

dmlond (changes requested):

app/api/dds/v1/auth_providers_api.rb
dmann: That's the max number of results passed back from ldap, which is set here:

https://github.com/Duke-Translational-Bioinformatics/duke-data-service/pull/1253/files#diff-a89a3c8956854e179524fb7c9c35a0e9R58

It could be an IdentityProvider attribute, but it's specific to LDAP. Also, this line is a big fat hack. Actually, that could be said of most of the code associated with this PR. This is definitely a first iteration to see if the stated theory will increase performance. If it turns out that there are performance gains, I will refactor the code to be less kludgey.

app/models/ldap_identity_provider.rb
dmann: :man_shrugging:

app/api/dds/v1/auth_providers_api.rb
dmlond: ok, that makes sense :)

app/models/ldap_identity_provider.rb
dmann: I could... But why?

dmlond: simpler? no results << nil.

dmann: Oh, no. I see what you're doing. That would end up breaking the pagination in the request, which takes the results of `#affiliates` and passes it to `Kaminari.paginate_array`, then to the `paginate` method.

dmlond: I'm not sure I follow, but it doesnt matter since this is iteration one anyway.

dmlond (approved):
seems like a good first pass

PR #1263, opened 2018-12-10
DDS-1170-auth_service_serializer_enhancements

dmlond:
added base_uri and login_response_type to the serializer
did not add client_id as this may be sensitive


dmann (approved):
Looks good :shipit:

PR #1265, opened 2018-12-26
DDS-1171 job to initialize chunked upload

dmann:
- [x] Add Upload#initialize_storage method
- [x] Create background job to call Upload#initialize_storage
- [x] Perform job after Upload#create
- [x] Update S3StorageProvider#chunk_upload_url to throw exception when Upload#multipart_upload_id is not set
- [x] Raise ConsistencyException when attempting to create a chunk for an upload that is not ready
- [x] Upload#ready_for_chunks? method
- [x] Upload serializer status hash includes :ready_for_chunks and :is_consistent

### DIFF
--- a/app/models/upload.rb
+++ b/app/models/upload.rb
@@ -10,6 +10,7 @@ class Upload < ActiveRecord::Base
   has_many :fingerprints
 
   before_create :set_storage_container
+  after_create :initialize_storage
 
   accepts_nested_attributes_for :fingerprints
 
--- a/app/models/upload.rb
+++ b/app/models/upload.rb
@@ -69,6 +69,10 @@ class Upload < ActiveRecord::Base
     )
   end
 
+  def ready_for_chunks?
+    storage_provider.chunk_upload_ready?(self)
+  end
+
   def complete
     transaction do
       self.completed_at = DateTime.now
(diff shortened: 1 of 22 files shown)


dmlond (approved):
👍

dmlond (approved):
💯

app/models/upload.rb
```diff
@@ -74,7 +74,7 @@ def ready_for_chunks?
   end
 
   def check_readiness!
-    ready_for_chunks? or raise ConsistencyException, 'Upload is not ready'
+    ready_for_chunks? || raise(ConsistencyException, 'Upload is not ready')
```
dmlond: raise(ConsistencyException, 'Upload is not ready') unless ready_for_chunks?

dmann: :man_shrugging: :potato: :tomato:

dmann: I can change it to use `unless` if you feel that it'll be more clear.

dmlond: I do think it is a little more clear 👓

dmlond (approved):
looks great

app/models/upload.rb
dmann: Sooo, that caused a test to fail because the `#check_readiness!` method now returns `nil` instead of a truthy value... I think I'm going to change it back to using `||` now, unless you have any objections.

dmlond: I still think its more readable as a guard clause
```
raise(ConsistencyException, 'Upload is not ready') unless ready_for_chunks?
true
```

dmlond (approved):
solid

PR #1266, opened 2019-01-03
DDS-1172 Fix CircleCi Bundler issue

dmann:
The call to `gem install bundler` is failing with:
ERROR: Error installing bundler:
The last version of bundler (>= 0) to support your Ruby & RubyGems was 1.17.3. Try installing it with `gem install bundler -v 1.17.3`
bundler requires RubyGems version >= 3.0.0. The current RubyGems version is 2.7.6. Try 'gem update --system' to update RubyGems itself.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -60,7 +60,7 @@ jobs:
             0.0.0.0       rabbitmq.local
             ' | tee -a /etc/hosts
 
-      - run: gem install bundler
+      - run: gem install bundler -v '~> 1.17.0'
       - run:
           name: Which bundler?
           command: bundle -v


.circleci/config.yml
```diff
@@ -60,7 +60,7 @@ jobs:
             0.0.0.0       rabbitmq.local
             ' | tee -a /etc/hosts
 
-      - run: gem install bundler
+      - run: gem install bundler -v '~> 1.17.0'
```
dmlond, line 63: does this also update rubyGems to >= 3.0.0?

dmann: No. It just specifies the version of `bundler` to install, which is the same version that we've been using. If we update RubyGems, then we can go back to not specifying the `bundler` version. I decided to go this route because the version of RubyGems in the `ruby:2.3.6` docker image is `2.7.6` and updating that would probably mean updating ruby, which is much more involved than setting the `bundler` version.

dmlond: ok.

dmlond (approved):
👍

PR #1272, opened 2019-01-16
Deploy v1.10.2 to ua_test

dmann:
Adding support for an S3 storage provider and the ability to choose a storage provider when initializing an upload for a project (`POST /projects/:id/uploads`).


dmlond (approved):
👍

PR #1273, opened 2019-01-22
DDS-1177 Env toggle message log worker indexing

dmann:
Skip message log indexing when MESSAGE_LOG_WORKER_INDEXING_DISABLED is set


dmlond (approved):
👍

PR #1277, opened 2019-02-05
DDS-1181 retry jobs with exponential backoff

dmann:
Reconfiguring sneakers to use the ExponentialBackoffHandler from https://github.com/alphasights/sneakers_handlers

The exchanges and queues created by this handler are significantly different from the MaxRetry handler. As a result, the ErrorQueueHandler class and the `queue:errors` rake tasks are deprecated until they can be rewritten.


dmlond (approved):
looks good

PR #1279, opened 2019-02-07
Deploy v1.10.8 to ua_test

dmann:
Retry jobs with exponential backoff
DEPLOYMENT NOTE: Spin down all dynos before deploying. All queues must be deleted before dynos can be spun back up.


dmlond (approved):
nice Deployment Note
👍

PR #1282, opened 2019-02-08
Deploy v1.10.x to production

dmann:
- Adding support for an S3 storage provider and the ability to choose a storage provider when initializing an upload for a project (POST /projects/:id/uploads).
*DEPLOYMENT NOTE: Once deployed, create the S3StorageProvider with `rake storage_provider:create STORAGE_PROVIDER_TYPE=s3`*
- Updating gems and addressing deprecation warnings
- Retry jobs with exponential backoff
*DEPLOYMENT NOTE: Spin down all dynos before deploying. All queues must be deleted before dynos can be spun back up.*
- Minimum chunk number validation


dmlond (approved):
👍

PR #1283, opened 2019-02-19
DDS-1182 non-chucked upload api design

dmann:


api_design/DDS-1182-nonchucked_upload_api_design.md
```diff
@@ -0,0 +1,157 @@
...
+
+`POST /projects/{id}/uploads`
+
+###### Request Properties
+
+- *name (string, required)* - The name of the client file to upload.
+- *content_type (string, optional)* - Valid content type per [media types](https://en.wikipedia.org/wiki/Internet_media_type).
+- *size (number, required)* - The size in bytes of entire file (computed by client).
+- *storage_provider.id (string, optional)* - The unique id for a storage provider.
+- *non_chunked (boolean, optional)* - When true, a signed upload url is returned with the payload
```
dmlond: since it is optional, it should state that the default is false

dmlond (changes requested):
just two little changes. This looks good.

api_design/DDS-1182-nonchucked_upload_api_design.md
dmlond: also, POST /projects/{id}/uploads and GET /uploads/{id} should include an example of a non_chunked: false response to make it clear what to expect.

dmann: > since it is optional, it should state that the default is false

Done.

dmann: > also, POST /projects/{id}/uploads and GET /uploads/{id} should include an example of a non_chunked: false response to make it clear what to expect.

I'd rather keep the examples in this design doc to what is changing/being added. I'll add in a sentence or two explaining that the proposed changes are backwards compatible and responses will remain consistent with the previous version when `non_chunked` is not set.

dmann: @dmlond  Ok, I added a few words about backwards compatibility. Check out my changes and let me know what you think. Thanks!

dmlond (approved):
This looks great!

johnbradley:
I like the idea of reducing the number DukeDS API calls. 👍 

You could make the new boolean attribute `chunked` and defaulted to `true`. When `non_chunked = false` the double negative seems confusing, but that just may be me.

Currently DukeDSClient retries 404 responses w/`resource_not_consistent` for 
- `POST /projects/{id}/uploads`  
- `PUT /uploads/{id}/chunks`
For non chunked upload it seems like we will only retry `POST /projects/{id}/uploads`.
Is there a need to call `GET /uploads/{id}`? Would this refresh the external url if it expired?

dmann:
> You could make the new boolean attribute `chunked` and defaulted to `true`. When `non_chunked = false` the double negative seems confusing, but that just may be me.

I think you're right about removing the double negative. I'll change the new boolean attribute to `chunked` as you suggested.

> For non chunked upload it seems like we will only retry `POST /projects/{id}/uploads`.

Yup, that's the only one you have to worry about retrying.

> Is there a need to call `GET /uploads/{id}`? Would this refresh the external url if it expired?

That's right, you'd call `GET /uploads/{id}` to refresh the signed upload url if your current one has expired.

johnbradley:
LGTM: 👍

PR #1286, opened 2019-03-04
DDS-1187 configure s3 bucket addressing

dmann:
Adding a field to allow S3StorageProvider use virtual host style bucket addressing instead of path style bucket addressing. For backwards compatibility, path style is the default. If it's determined that virtual host style should be the default, that can be addressed in a future PR.


dmlond (approved):
👍

PR #1297, opened 2019-03-21
Sync Bot

dmlond:
this activates bin/bot/sync_pr.rb
requires an org context sync_bot (already defined) with environment required by sync_pr.rb
will automatically create a PR from production to develop each time something is merged into production


.circleci/config.yml
```diff
@@ -1,5 +1,47 @@
...
+      - image: ruby:2.3.6
+        environment:
+          BUNDLE_JOBS: 3
+          BUNDLE_RETRY: 3
+          BUNDLE_PATH: vendor/bundle
+          RAILS_ENV: develop
+    steps:
+      - checkout
+      - run:
+          name: install dockerize
```
dmann: It doesn't look like `sync_bot` actually use the `dockerize` command, am I missing something?

dmlond: probably not. just copypasta :)

dmann (approved):
Looks good! :shipit:

PR #1305, opened 2019-04-02
DDS-1195 Purge non-chunked uploads issue

dmann:
Fixing a bug that prevented non-chunked uploads from being purged from Swift storage providers.


dmlond (approved):
💯

PR #1312, opened 2019-04-09
DDS-1190 code coverage

dmann:
- Use SimpleCov to calculate code coverage
- Adding `coverage` workflow to CI to merge parallel coverage results and store in job's artifacts

### DIFF
--- a/.circleci/report_coverage.rb
+++ b/.circleci/report_coverage.rb
@@ -1,0 +1,14 @@
+#!/usr/bin/env ruby
+
+require 'simplecov'
+
+base_dir = ENV.fetch("COVERAGE_BASE_DIR", "./coverage_results")
+file_pattern = ENV.fetch("COVERAGE_NAME_PATTERN", ".resultset*.json")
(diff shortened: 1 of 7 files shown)


.circleci/report_coverage.rb
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env ruby
```
dmlond, line 1: I dont know if we need to be consistent, but in other projects where I have created executables for use in CI, I have used `bin/ci` to store the executables.

dmlond (changes requested):

.circleci/report_coverage.rb
dmann: Currently the CI executables for this project are in the `.circleci` directory. I'm fine with refactoring the directory structure for the project to follow a more generic standard, but I think that's out of the scope of this PR. Maybe make a card in Jira and we can address this issue, as well as the directory structure as a whole.

dmlond: ok. that's fine. I just wanted to bring it up before it was too late

dmlond (approved):
💯

PR #1313, opened 2019-04-09
DDS-1197 Address code coverage issues

dmann:
- Allow SimpleCov exit behaviour to be configured via ENV
- Upload coverage results to codecov.io via CI


dmlond (approved):
looks cool

PR #1322, opened 2019-05-02
Heroku-16 Stack Migration

dmlond:
updated to use ruby 2.6.3 and use the debian stretch based ruby:2.6.3-stretch base image in our local docker image

Debian stretch is the upstream of Ubuntu 16.4, which is what the Heroku-16 stack is based upon. This should make our local docker image similar to the heroku environment


dmann (approved):
:shipit:

Issue #1329, opened 2019-05-30
Additional versions of files not updating last_updated_on


johnbradley:
Additional versions of files uploaded via __DukeDSClient__ do not have their `audit.last_updated_on`, `audit.last_updated_by` fields updated. Files uploaded by `duke-data-service-portal` do not have this problem. I have been unable to determine what the difference might be. Any feedback on what I might be missing is appreciated.

An example file that doesn't have last_updated_on/last_updated_by filled in but has multiple versions is https://dataservice.duke.edu/#/file/c8596497-6873-40a8-975a-b4d119cb3f08.
This issue can be easily reproduced running the ddsclient upload command to upload a file, changing the file and uploading the file again. For users this problem is seen in `duke-data-service-portal` where the the last updated date is wrong.

Original issue in DukeDSClient: https://github.com/Duke-GCB/DukeDSClient/issues/239

dmann:
It looks like the portal always sends `label` (even when it's blank) in the payload when it calls `PUT /api/v1/files/{id}`, which can trigger an update for the file's audit. Adding a file version by itself isn't triggering the file audit update. This explains the inconsistency that you're seeing between the two clients.

I'll work on correcting the file version bug and try to get a PR to production in time for next week's maintenance window.

dmann:
A fix has been deployed to production. If you find that the issue still exists, feel free to reopen this issue or create a new one. Cheers! :beers:

PR #1330, opened 2019-06-05
DDS-1212 create DataFile audit for new file_versions

dmann:
- Adds an `etag` attribute to `containers` table
- Adds callback to `DataFile` to set `etag` whenever the object itself or `file_versions` association changes. Updating the `etag` will trigger an audit for `DataFile`, which should address issue #1329.


dmlond (approved):
👍

PR #1378, opened 2021-08-25
Deploy v1.11.21 to Production

dmann:
- BUG FIX: update s3 gem
- Adding SingleBucketS3StorageProvider model which uses a single S3 bucket to store Uploads. Objects are stored will have keys prefixed with the Upload's Project id, like `project_id/upload_id`.
- The `storage_provider:create` rake task has been updated to accept a `STORAGE_PROVIDER_TYPE` of `single_bucket_s3` and a set of env vars prefixed with `SINGLE_BUCKET_S3_`.


dmlond (approved):
ship it 🤞

